PR #10, opened 2016-02-22
TCOMP-51 Make changes in a form visible until canceled

francisu:


daikon/src/main/java/org/talend/daikon/properties/Property.java
```diff
@@ -141,11 +142,16 @@ public int getIntValue() {
         return Integer.valueOf(String.valueOf(value));
     }
 
-    /**
-     * @return Cast the stored value to a Calendar.
-     */
     public Calendar getCalendarValue() {
```
sgandon: Francis, can we talk about this, I think this should not be part of this class. If you really think the API should provide some converters which which I don't think they should be part of this api, because you expect the called to know about the content type and therefore they can do the convertion them self or through some external helper class. 
But you can't have throw IllegaStateException like that in so many method expecting people will not call them.
I strongly suggest we create a convert external class if you really want to provide anything, but I think this is the component designer to handle it's how type.

daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
```diff
@@ -83,7 +83,7 @@
 
     private boolean cancelable;
 
-    private Map<String, Object> cancelableValues;// stores the value temporarily in case the are canceled
+    private Map<String, Object> originalValues;// stores the value temporarily in case the are canceled
```
sgandon: I think this will increase the size of the payload for conveying data that will not being used by the client. If we think about it the original value are already stored in the backend, so if you want to use them don't you think we could get them from this backend-storage whenever we need them ?

daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
```diff
@@ -257,13 +257,18 @@ public Widget getWidget(Class<?> childClass) {
         return null;
     }
 
+    public Form getChildForm(String child) {
```
sgandon: here also you do a cast that will fail most of the time as child content are not always forms.

francisu: This is used only if you are expecting a Form. It's currently only used by the tests.

daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
francisu: I think doing something like that would complicate things. Right now, everything is stored in the ComponentProperties and there is no actual back-end storage. Let's see if we actually have a performance problem before we do something additional. And this only applies to forms that are cancelable (like a dialogue with a cancel button).

daikon/src/main/java/org/talend/daikon/properties/Property.java
francisu: I'm happy to talk about this. I think the methods for getting the values and the types of values allowed are not fully thought through and should be. Maybe instead we should have a getDateValue()? I'm inclined to agree with you that we should not do conversions. This was done without much thought. Changing it is welcome.

PR #11, opened 2016-02-25
TCOMP-103 : added a way to assign recursively nested properties

sgandon:
I have also added a test util method to generated the list of compatible nested properties for setting then during development.


daikon/src/test/java/org/talend/daikon/properties/PropertiesTest.java
```diff
@@ -207,16 +208,6 @@ public void testGetProps() {
...
-    // @Test
-    // public void testGetPropFields() {
-    // TestProperties tProps = (TestProperties) new TestProperties("test").init();
-    // List<String> fieldNames = tProps.getPropertyFieldNames();
-    // System.out.println(fieldNames);
-    // assertEquals(11, fieldNames.size());
-    // assertTrue(tProps.userId == tProps.getPropertyByFieldName("userId"));
-    // assertTrue(tProps.nestedProps == tProps.getPropertyByFieldName("nestedProps"));
-    // }
-
```
francisu: Why was this test removed?

sgandon: For 2 reason, it was already commented and it tests a method that does not exists anymore ;).

PR #12, opened 2016-02-25
Update version to 0.3.1-SNAPSHOT and inline avro.

RyanSkraba:
I need an OSGi expert to take a look at this -- I spent more than a few hours trying to get daikon to work in the studio with Avro and finally broke down to this.

In short, avro can't be delivered as a library for many of the components in the Studio because of a conflict between jackson 1 and jackson 2.

### DIFF
--- a/daikon-spring/pom.xml
+++ b/daikon-spring/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.talend.daikon</groupId>
 		<artifactId>daikon-parent</artifactId>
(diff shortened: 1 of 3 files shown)


RyanSkraba:
Hello!  Do not merge this pull request!  Comments only :)

sgandon:
I would suggest you only embed avro and it's dependencies.
{code}
<Embed-Dependency>*;artifactId=avro|jackson-core-asl|jackson-mapper-asl|paranamer|snappy-java|commons-compress|xz</Embed-Dependency>
{code}
Can you try that ?

PR #13, opened 2016-03-16
remove Form.widgetList and only keep Form.widgetMap.

sgandon:
the Form used to have a list of widget and a map with the same set of widgets.
To simplify the class, the understanding and the serialization I removed the list and kept only the map which I changed to an LinkedHashMap to have keep the insertion order that is required for the widget list.

### DIFF
--- a/daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
+++ b/daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
@@ -12,9 +12,9 @@
 // ============================================================================
 package org.talend.daikon.properties.presentation;
 
-import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
-import java.util.List;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 import org.talend.daikon.NamedThing;
@@ -72,9 +72,7 @@ public class Form extends SimpleNamedThing implements ToStringIndent {
     @JsonBackReference
     protected Properties properties;
 
-    protected Map<String, Widget> widgetMap;
-
-    protected List<Widget> widgets;
+    protected LinkedHashMap<String, Widget> widgetMap;
 
     private int currentRow;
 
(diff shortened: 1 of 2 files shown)


francisu:
+1

daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
```diff
@@ -223,22 +222,27 @@ public void setVisible(boolean visible) {
...
     }
 
-    private void fill(Widget widget) {
+    /**
+     * DOC sgandon Comment method "getWidgetName".
+     * 
+     * @param widget
+     * @return
+     */
+    public String getWidgetName(Widget widget) {
```
francisu: Please remove this useless DOC comment.

sgandon:
This was fixed, thanks for pointing this out.
I have updated the component talend_formatter to not have this default javadoc comment on new methods. 
I'll merge this myself if you don't mind because there is a slight change in the Form APIs.
```
    public List<Widget> getWidgets() {
        return widgets;
    }
```
was changed to 
```

  public Collection<Widget> getWidgets() {
        return widgetMap.values();
    }
```
So this might impact the Studio (or not depending on how they used it).
I need to build the Studio with this change to check.

sgandon:
merged manually.

francisu:
This change broke serialization compatibility with existing properties. There are a couple of issues here: 1) we should not persist the forms stuff since it will never be used when reading the properties from a store (however when communicating on the wire of course we need to serialize it. 2) we need a means of dealing with this sort of incompatibility in the future. I wrote a wiki page about this.

sgandon:
1) This is strange because your first point should be already implemented for a while : https://github.com/Talend/daikon/blob/master/daikon/src/main/java/org/talend/daikon/properties/Properties.java#L334
See the toSerialized, it does set the forms to null before serializing.

francisu:
Yes, I remember this, but when build a studio with these changes, the deserialiation failed because of this change. So maybe there is something not understood here. But the main point I bring it up is that we have to be careful of any changes that affect the serialized version. We also need to make sure we have a means of versioning/checking for wire compatibility so that we make sure the front end and back end (REST service) match correctly. Perhaps this can be controlled with just the version number of the overall service (from the components-api POM), but I'm not sure this is granular enough.

sgandon:
indeed but have not yet fixed apis. The version is already checked from the current bundle version of the api and daikon they use. I we break some APIs we just have to augment the version of the pom and then the client decides to use it or not.
This problem is more a migration problem to me that we do not address yet.

PR #14, opened 2016-03-18
TCOMP-87-Dynamic

bchen-talend:


daikon/src/main/java/org/talend/daikon/avro/DynamicTypeFactory.java
```diff
@@ -0,0 +1,55 @@
+package org.talend.daikon.avro;
+
+import org.apache.avro.LogicalType;
+import org.apache.avro.LogicalTypes;
+import org.apache.avro.Schema;
+
+/**
+ * Created by bchen on 16-3-18.
+ */
+public class DynamicTypeFactory {
```
francisu: Can you remove this comment? We generally don't need these (source control will say who created what).

daikon/src/main/java/org/talend/daikon/avro/util/AvroUtils.java
```diff
@@ -84,15 +83,13 @@ public static Schema unwrapIfNullable(Schema schema) {
...
-        lt.addToSchema(fs);
-        return field;
+    public static boolean hasDynamicField(Schema schema) {
+        for (Schema.Field field : schema.getFields()) {
+            if (DynamicTypeFactory.isDynamic(field.schema())) {
+                return true;
+            }
+        }
+        return false;
     }
```
francisu: You should remove this method from AvroUtils, as it's the same as your DynamicTypeFactory. We need to make sure to do only one thing in one place.

PR #17, opened 2016-03-23
remove the Dynamic from component framework, and only handle it in Talend6OutgoingSchemaEnforcer

bchen-talend:


daikon/src/main/java/org/talend/daikon/avro/util/AvroUtils.java
```diff
@@ -84,15 +83,60 @@ public static Schema unwrapIfNullable(Schema schema) {
...
-        return schema.getType() == Type.BYTES && schema.getLogicalType().getName().equals(SchemaConstants.LOGICAL_DYNAMIC);
+    /**
+     * Schema don't support overwrite property's value, so have to clone it then put the new value
+     *
+     * @param schema
+     * @param key
+     * @param value
+     * @return schema with the new value for the property: key
+     */
+    public static Schema setProperty(Schema schema, String key, String value) {
```
RyanSkraba: Just a quick note -- you *can* add a new property to a schema without any problems, so this method is unnecessary if we limit the developer to ADDING "include-all-fields" (and never changing or removing it).

But I think this is a better way of doing it.  I am not surprised that it is so difficult to modify a Schema in Avro...

daikon/src/main/java/org/talend/daikon/avro/util/AvroUtils.java
```diff
@@ -84,15 +83,60 @@ public static Schema unwrapIfNullable(Schema schema) {
...
+                Map<String, Object> props = schema.getObjectProps();
+                for (String propKey : props.keySet()) {
+                    if (propKey.equals(key)) {
+                        newSchema.addProp(key, value);
+                    } else {
+                        newSchema.addProp(propKey, props.get(propKey));
+                    }
+                }
+            } else {//FIXME for other type
+                throw new RuntimeException("Not support this type " + newSchema.getType() + " now, need to implement");
```
RyanSkraba: This is fine for me for now.  It will need to be fully implemented for fully hierarchical structures.

daikon/src/main/java/org/talend/daikon/avro/util/AvroUtils.java
```diff
@@ -84,15 +83,60 @@ public static Schema unwrapIfNullable(Schema schema) {
...
-        Schema fs = field.schema();
-        lt.addToSchema(fs);
-        return field;
+    /**
+     * @param schema
+     * @return the value of the property include-all-fields, false if there is no this property
+     */
+    public static boolean isIncludeAllFields(Schema schema) {
+        String prop = schema.getProp(SchemaConstants.INCLUDE_ALL_FIELDS);
+        if (prop == null || !Boolean.valueOf(prop)) {
```
RyanSkraba: Simpler:
return prop != null && Boolean.valueOf(prop);

RyanSkraba:
This looks good to me!  I am checking the related pull requests:

https://github.com/Talend/components/pull/37/files
https://github.com/Talend/tcommon-studio-se/pull/341/files
https://github.com/Talend/tdi-studio-se/pull/263/files

PR #19, opened 2016-03-25
TCOMP-69: Change Property defualtVlaue type form String to Object.

jzhao-talend:


sgandon:
Thank for this pull request jzhao, but we have had some talk with Francis and we think that default value could be removed. Could you do that instead please.

jzhao-talend:
@sgandon 
Yes. you are right.
In component side, we can replace **setDefaultValue** with **setValue**
If defautValue isn't used in other side, it can be removed.
I would push another commit in component side.

jzhao-talend:
Have replaced "setDefaultValue" with "setValue" in components PR.

PR #21, opened 2016-03-30
complete built-in Avro type

bchen-talend:


daikon/src/main/java/org/talend/daikon/avro/SchemaConstants.java
```diff
@@ -6,6 +6,11 @@
 public class SchemaConstants {
 
     /**
+     * If Schema.Type can't represent this java type, then use JAVA_CLASS_FLAG as the property key and the real class name as the value
+     */
+    public static final String JAVA_CLASS_FLAG = "java-class";
```
RyanSkraba: We should use SpecificData.CLASS_PROP instead of defining a new constant here.

RyanSkraba:
Other than the comment on the JAVA_CLASS versus SpecificData.CLASS_PROP constant, this looks good to me!

daikon/src/main/java/org/talend/daikon/avro/SchemaConstants.java
RyanSkraba: Don't change this :)  There is an issue with our Avro bundle in the studio that will prevent the SpecificData.CLASS_PROP from being used for now.

PR #23, opened 2016-04-08
TCOMP-143 Support obtaining properties from upstream components

francisu:


francisu:
Should not be merged until the corresponding branch in components is done.

sgandon:
See my comment in https://jira.talendforge.org/browse/TCOMP-143, I don't think we need this right now and for the given requirement we could solve this using input/output schemas from input/output connectors.

daikon/src/main/java/org/talend/daikon/properties/DesignerService.java
```diff
@@ -0,0 +1,7 @@
+package org.talend.daikon.properties;
+
+/**
+ *
+ */
+public interface DesignerService {
```
sgandon: How come this interface is empty ?

francisu: It's intended to be implemented by a sub-interface in Components (which will have what was in Repository and a new method to ask about connected components). Daikon does not need/support any of the methods. But we need the interface because it's a parameter in one of the methods that Daikon has.

sgandon:
I think this branch can be removed, as we found another way to solve this.

sgandon:
@francisu can we  close this ?

francisu:
See https://jira.talendforge.org/browse/TDKN-107

PR #26, opened 2016-04-19
[TDKN-22] add isKey on SchemaConstant

bchen-talend:
[TDKN-22] add isKey on SchemaConstant


RyanSkraba:
This looks good to me and can be merged!  Can you add javadoc to the SchemaConstants interface to describe where this should be used.  I think it is only set to true when the original data source has an index on that column?  Or just a primary key index?

If you want to think about and propose a better mechanism for database generic or specific annotations on columns, I'd love to hear it!

PR #27, opened 2016-04-19
[TDKN-23] Remove defaultValue from Property

bchen-talend:
[TDKN-23] Remove defaultValue from Property

### DIFF
--- a/daikon/src/main/java/org/talend/daikon/properties/Property.java
+++ b/daikon/src/main/java/org/talend/daikon/properties/Property.java
@@ -12,20 +12,12 @@
 // ============================================================================
 package org.talend.daikon.properties;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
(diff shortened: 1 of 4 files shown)


francisu:
Will you make a corresponding components request that goes with this?

bchen-talend:
@francisu I checked salesforce component, no change needed, but may effect studio team maybe, need check

francisu:
Almost certainly should be no impact with the studio, but good to check anyway. Let me know when you have checked and Seb or I will merge this.

nrousseau:
Is it possible to set the old method of getDefault() / setDefault() as deprecated simply for now?
Seems much better rather than have a synchronized task

sgandon:
if you deprecate them , they will go away eventually, then you'll have to check the studio anyway, so I'd say we do it now since the Studio is our only client right now.

sgandon:
One comment though on the naming, since it is not the default value anymore because we will never be able to get it later, I would suggest to rename the parameters *initialValue* instead of *defaultValue*.

bchen-talend:
Ok, @sgandon I will use initialValue as the name.
Do you think we can remove the following method directly?
```
@Deprecated
Property newInteger(String name, String defaultValue)
```

sgandon:
nope this is a helper method that is useful, why would you add @Deprecated it ?

PR #29, opened 2016-04-28
Feature/tcomp 175 runtime output records

RyanSkraba:
For https://jira.talendforge.org/browse/TCOMP-175, we need to make sure that output components are getting an IndexedRecord with the "actual" schema of the discovered data (including dynamic columns).  

It looks like we can do this more efficiently with the same pattern as enforcing outgoing schemas, so I've added the additional helper class, which *should not* be used by other components (and will certainly be moved to DI code with the others from TCOMP-176).

This goes along with https://github.com/Talend/tdi-studio-se/pull/370, but to be discussed here!


sgandon:
I am fine with it, please can you merge it?

PR #30, opened 2016-04-29
TUP-4404 : added a IS_VISIBLE tagged value to Property set when widge…

sgandon:
…t changes it's visibility.
@francisu I did set the TaggedValue of the Widget's Property when the visibility changes.
I just called is IS_VISIBLE because it is what it represents for now. I did not want to add some runtime_visibility who sound approximative to me.

### DIFF
--- a/daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
+++ b/daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
@@ -177,7 +177,7 @@ public class Widget implements ToStringIndent {
             if (hidden)
                 prop.addFlag(Property.Flags.HIDDEN);
             else
-                prop.getFlags().remove(Property.Flags.HIDDEN);
+                prop.removeFlag(Property.Flags.HIDDEN);
         }
         return this;
     }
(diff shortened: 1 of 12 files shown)


daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
```diff
@@ -102,7 +102,7 @@
      */
     private int order;
 
-    private boolean visible = true;
+    private boolean hidden;
```
sgandon: I suggest you set some default value so that we know what it is. Especially here I think we should have false.

daikon/src/test/java/org/talend/daikon/avro/SchemaTest.java
```diff
@@ -0,0 +1,26 @@
+package org.talend.daikon.avro;
```
sgandon: I don't understand why did you add an Avro schema test in this pull request. This is unrelated to the widget and property hidden.

daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
francisu: The default value in Java for something that is not initialized is either null, 0, or false. I don't believe in setting things explicitly to something they already are.

daikon/src/test/java/org/talend/daikon/avro/SchemaTest.java
francisu: True, it was related to something else (turned out to be a false alarm). But are you going to complain about more tests?

daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
sgandon: This way you make easier for some that read the code to understand the initial value and you also tell the the reviewer that you did no forget to set it by mistake.

PR #31, opened 2016-05-02
TDKN-26 Rename the Talend6 Schema constants to refer to 'di' instead …

RyanSkraba:
…of 'talend6'

This *might* break jobs that were created with Talend 6.2.0RC1 and before, but should be done before Talend 6.2.0GA.

I was unable to test the impact, so we might want to wait until after the GA and do the conversion with a mapping table between 6.2.0 and 6.2.1.


francisu:
If we put this in, it will go into Daikon version 0.8.0 (and thus 6.2.0). Should we go ahead?

sgandon:
I believe so.

francisu:
OK then.

PR #32, opened 2016-05-03
Tdkn 35 make property generic

sgandon:
I changed the Property class to use generic type and removed the inner enum Type.
Please can someone review that.


francisu:
Looks OK for me, though we should put it in 0.9.0, so you should increase the version number before the merge.

sgandon:
yes I shall update it to 0.10.0-SNAPSHOT because master is now at 0.9.0-SNAPSHOT.

PR #42, opened 2016-06-06
TCOMP-17 Support persistent object migration

francisu:


daikon/pom.xml
```diff
@@ -77,6 +77,11 @@
 			<groupId>com.cedarsoftware</groupId>
 			<artifactId>json-io</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<version>4.11</version>
```
sgandon: please add 
            <scope>test</scope>
otherwise junit will be part of the runtime dependencies for daikon.

daikon/src/main/java/org/talend/daikon/properties/Properties.java
```diff
@@ -89,30 +90,12 @@
...
-
-        public static synchronized <T extends Properties> Deserialized<T> fromSerialized(String serialized,
+        public static synchronized <T extends Properties> SerializerDeserializer.Deserialized<T> fromSerialized(String serialized,
                 Class<T> propertiesclass) {
-            return fromSerialized(serialized, propertiesclass, null);
+            return PropertiesImpl.fromSerialized(serialized, propertiesclass);
         }
     }
 
-    /**
```
sgandon: Why did you remove this, this was required by the Studio to avoid properties initialization and layout setup to fail if the stored properties needs to be converted to the right property type ?

daikon/src/main/java/org/talend/daikon/properties/PropertiesImpl.java
```diff
@@ -56,49 +54,37 @@
...
-        // this set the proper classloader for the JsonReader especially for OSGI
-        ClassLoader originalContextClassLoader = Thread.currentThread().getContextClassLoader();
-        try {
-            Thread.currentThread().setContextClassLoader(Properties.class.getClassLoader());
-            d.properties = (T) JsonReader.jsonToJava(serialized);
-            if (d.properties instanceof PropertiesImpl) {
-                ((PropertiesImpl) d.properties).handlePropEncryption(!ENCRYPT);
-            } // else we nothing to be done
-            if (postSerializationSetup != null) {
-                postSerializationSetup.setup(d.properties);
```
sgandon: same as above, this is important to setup evaluators here before any initialization.

daikon/src/main/java/org/talend/daikon/serialize/DeserializeDeletedFieldHandler.java
```diff
@@ -0,0 +1,25 @@
...
+public interface DeserializeDeletedFieldHandler extends DeserializeMarker {
+
+    public static final String FIELD_DELETED_PREFIX = "fieldDeleted_";
+
+    /*
+     * Called when a field is present in the serialized data, but is not present in the object being deserialized.
+     * 
+     * @param value the value of the deleted field in the old object
+     * @return true if the object is considered to have migrated (it was modified from the serialized version)
+     */
```
sgandon: I already mentionned that I think marker interface should be seldomly used. Here we could have a single method like
`        fieldDeleted(String field name, Object value)`
Actually the marker pattern has been usually abandoned to use annotation and that would be my recommendation, define some method annotation like this
@REMOVED_FIELD("field1")
boolean anyMethodName(Object value)

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
```diff
@@ -0,0 +1,156 @@
...
+        Deserialized<T> d = new Deserialized<T>();
+        d.migration = new MigrationInformationImpl();
+        ClassLoader originalContextClassLoader = Thread.currentThread().getContextClassLoader();
+        try {
+            // OSGi requires the the classloader for the target class
+            Thread.currentThread().setContextClassLoader(serializedClass.getClassLoader());
+
+            final Map<PostDeserializeHandler, Integer> postDeserializeHandlers = new HashMap<>();
+
+            JsonReader.JsonClassReaderEx reader = new JsonReader.JsonClassReaderEx() {
```
sgandon: I would rather have those 2 anonymous classes be real classes for more readability.

daikon/src/test/java/org/talend/daikon/properties/PropertiesTest.java
```diff
@@ -378,22 +384,19 @@ public void testTaggedValue() {
...
+            public boolean postDeserialize(int version, boolean persistent) {
+                date.setTaggedValue("foo", "bar" + version);
+                return false;
+            }
+        };
+        props.init();
         String s = props.toSerialized();
-        final AtomicBoolean setupCalled = new AtomicBoolean(false);
-        assertFalse(setupCalled.get());
         assertNotEquals("bar", props.date.getTaggedValue("foo"));
```
sgandon: here you should do a negative check for the same value "bar0" as the one you check after the deserialization.

daikon/src/test/java/org/talend/daikon/serialize/PersistenceTestObject.java
```diff
@@ -0,0 +1,93 @@
...
+    public PersistenceTestObject() {
+        inner = new PersistenceTestObjectInner();
+    }
+
+    public void setup() {
+        // Original values
+        string1 = "string1";
+        // string2 = "string2";
+        string3 = "string3";
+        inner = new PersistenceTestObjectInner();
```
sgandon: why this is create here and in the constructor ?

daikon/src/test/java/org/talend/daikon/serialize/SerializeDeserializeTest.java
```diff
@@ -0,0 +1,89 @@
...
+            + "\"inner\":{\"string1\":\"string1\",\"string2\":\"string2\","
+            + "\"innerObject2\":{\"string1\":\"string1\",\"string2\":\"string2\"}}}";
+
+    @Test
+    public void testSimple() {
+        PersistenceTestObject.testMigrate = false;
+        PersistenceTestObject pt = new PersistenceTestObject();
+        pt.setup();
+        String ser = SerializerDeserializer.toSerialized(pt, SerializerDeserializer.PERSISTENT);
+        System.out.println(ser);
```
sgandon: I think it would be best to use slf4j *Logger.debug()* instead of system.out.println, even for test case.
Thanks.

pom.xml
```diff
@@ -153,6 +165,7 @@
 			</activation>
 			<build>
 				<plugins>
+					<!--
```
sgandon: please do not remove the formatter check. If your code format is not compliant with the formatter definition, please run
`          mvn clean install -Dformat` 
this will format all the code to be compliant.

daikon/pom.xml
francisu: I removed the dependency, it should not be  there.

daikon/src/main/java/org/talend/daikon/properties/Properties.java
francisu: I did not know of the studio use. I can put it back in. This PR uses a single post serialization mechanism.

daikon/src/main/java/org/talend/daikon/serialize/DeserializeDeletedFieldHandler.java
francisu: I'm OK with the annotation style; the reason that a generic method was not specified was that it is much easier with the implementation (and specification) to provide the field name declaratively (rather than having it in the code). I did try it in the code and did not like it. I will change it to use an annotation.

daikon/src/main/java/org/talend/daikon/properties/Properties.java
sgandon: Yes this has nothing to do with the migration/deserialization process, this is just a Properties issue that need to be handled.

pom.xml
francisu: This is a bigger issue. The eclipse formatter in intellij is not aligned with the formatter used by maven (I trued to used version 2.0.0 of the maven formatter but it could not get that version somehow). This is removed temporarily, but needs to somehow be fixed. The formatting needs to work the same for both Eclipse and Intellij. I'm using the formatting file in the tools repo.

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
francisu: OK

daikon/src/test/java/org/talend/daikon/serialize/PersistenceTestObject.java
francisu: A mistake. Fixed.

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
```diff
@@ -0,0 +1,156 @@
...
+
+                @Override
+                public Object read(Object jOb, Deque<JsonObject<String, Object>> stack, Map<String, Object> args) {
+                    JsonReader reader = Support.getReader(args);
```
francisu: Thanks!

francisu:
@sgandon I think I addressed all of your comments. I also added some additional simplification (I removed MigrationInfo* since it was really no longer needed since the version number is not required).

I added a PostDeserializationSetup interface which is similar to the PostSerializationSetup interface that was there (I removed that one, as I push this mechanism into the serialize package). I have a test that uses the property evaluator stuff just like the studio use case (I saw they are using this in 10 places for that). The studio changes to convert to this should be pretty quick (preserving the existing functionality). However, the studio will need to pay attention to the "migrated" flag now when an object is deserialized to give the right indication to the user (and perhaps resave the object). That will be extra work, but it's not required for them to take these changes.

daikon/src/main/java/org/talend/daikon/properties/PropertiesImpl.java
francisu: This has been replaced by the PostDeserializationSetup mechanism (see line 62)

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Returns a materialized object from a previously serialized JSON String.
+     *
+     * @param serialized created by {@link #toSerialized(Object object, boolean persistent)}.
+     * @param serializedClass the class of the object being deserialized
+     * @param persistent see {@link #PERSISTENT} and {@link #TRANSIENT}.
+     * @return a {@code Properties} object represented by the {@code serialized} value.
+     */
+    public static <T> Deserialized<T> fromSerialized(String serialized, Class<T> serializedClass, PostDeserializeSetup setup,
+            boolean persistent) {
```
sgandon: I don't understand this persistent flag, that only purpose is to being passed to callbacks. It does not seem to be used at all to change the way objects are persisted.
So I wonder what could be the use of it right now?

francisu: Right now it's not used, but it likely will be. For example, in the persistent case, we might want to not serialize any of the presentation objects. It's better to have the persistent/transient distinction be part of the API for both serialize/deserialize so that we don't have to change the API again.

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
```diff
@@ -0,0 +1,185 @@
...
+            Thread.currentThread().setContextClassLoader(originalContextClassLoader);
+        }
+        return d;
+    }
+
+    /**
+     * See {@link #toSerialized(Object, boolean)}
+     */
+    public static <T> String toSerializedPersistent(T object) {
+        return toSerializedPersistent(object);
```
sgandon: this looks like an infinite loop !

francisu:
@sgandon There is a problem with your change to omit the "forms" field (though the existing implementation was completely wrong as well, transient serialization was broken). Your change would only work for the top-level Properties object being serialized. I have added a call-back interface SerlizeFieldOmitter which will be used to say which fields can be omitted (it will have the persistent flag as an argument). I will also change the PR into json-io to better support this type of interface.

I'm going to go ahead and merge this now though, as the API changes are complete for the purposes of the studio and this will allow them to get going. The remaining work on this can be done without breaking anything.

PR #55, opened 2016-08-05
TUP-5101: Make Widget of TEXT_AREA available in new framework

ycbai:
TUP-5101: Make Widget of TEXT_AREA available in new framework

### DIFF
--- a/daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
+++ b/daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
@@ -99,6 +99,8 @@ public class Widget implements ToStringIndent {
      **/
     public static final String ENUMERATION_WIDGET_TYPE = "widget.type.enumeration";
 
+    public static final String TEXT_AREA_WIDGET_TYPE = "widget.type.textArea";
+
     /**
      * The row in the form where this property is to be presented. Starting with 1.
      */


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/60/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/60/)

nrousseau:
control is not standard right now for components, since it's only used in one component.
I think maybe no need to add to daikon right now (string of definition of widget can directly be used in the component)

francisu:
How then do we address this for the
One component that needs it?  I think it's a good idea to have this widget type, as it will come up with other components.

nrousseau:
This control is a multiline text.
Most of the components (DI) are using what we called "MEMO" for the multiline control.
Actually only one component use this "TEXT_AREA", and it's for a MDM component, which extends the memo type.
I'm not sure at all the TEXT_AREA should be the new standard right now...  And i'm still not sure about the code used in this control for the display.
cf: `param.getValue().toString().replace("\n", " ").replace("\r", " ")`
Might need the review of MDM to really explain and see if this control can be a standard one for every component or not.

lpw82gf:
MEMO doesn't support multi-line String definition as bellow:
The code change for TEXT_AREA is recommended by GGU of DI team.
I think the better way is change MEMO to support multi-line String definition, but to avoid impact, he advised to extend MEMO.
```
"{
    'select':{
        'from':[
            'ProductFamily'
        ],
        'fields':[
            {
                'field':'ProductFamily/Id'
            },
            {
                'field':'ProductFamily/Name'
            }
        ]
    }
}"
```

glzhou:
"MEMO" widget is not very convenient for user to construct a multi-line string nor easy to read, it requires to have lines concatenated with "+", to have the same input as above, we would have:

"{ " + 
    "'select':{" +
       " 'from':[" +
           " 'ProductFamily'" +
       " ]," +
        " 'fields':[" +
           " {" +
               " 'field':'ProductFamily/Id'" +
           " }," +
           " {" +
               " 'field':'ProductFamily/Name'" +
           " }" +
       " ]" +
   " }" +
"}"

nrousseau:
Merge after confirm with ggu that this could be a standard widget for the new framework

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/97/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/97/)

francisu:
:)

PR #57, opened 2016-08-22
TDKN-75 : add system properties isolation mechanism.

sgandon:
This adds a set a of classes for System properties isolation base the classloader and the thread context classloader.


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/75/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/75/)

sgandon:
I'll just change something in the *close()* method of  the *SandboxedInstance* so that it does not throw an exception, this just a pain to handle it'll be converted to a TalendException

daikon/src/main/java/org/talend/daikon/sandbox/SandboxInstanceFactory.java
```diff
@@ -0,0 +1,65 @@
...
+public class SandboxInstanceFactory {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SandboxInstanceFactory.class);
+
+    /**
+     * This will create a new class instance base on an URLClassLoader using the <code>classPathUrls</code> and the
+     * <code>parentClassLoader</code>.
+     * This instance will be embed in a {@link SandboxedInstance} so that it provides a SystemProperty isolation.
+     * 
+     * @param classToInstanciated name of the class to instanciate
```
RyanSkraba: Minor, unfortunate english: there is no 'c' in instantiate :(

daikon/src/main/java/org/talend/daikon/sandbox/SandboxInstanceFactory.java
```diff
@@ -0,0 +1,65 @@
...
+        if (classToInstanciated == null) {
+            throw new IllegalArgumentException("classToInstanciate should not be null");
+        }
+        ClassLoaderIsolatedSystemProperties isolatedSystemProperties = ClassLoaderIsolatedSystemProperties.getInstance();
+
+        try {
+
+            LOGGER.info("creating class '" + classToInstanciated + "'"); //$NON-NLS-1$ //$NON-NLS-2$
+
+            // the following classloader is closeable so there is a possible resource leak.
```
RyanSkraba: Just a note!  The Autoclosable is much nicer than the Thread-isolated version (which automatically reintegrated the thread at completion).  Good!

daikon/src/main/java/org/talend/daikon/sandbox/SandboxInstanceFactory.java
```diff
@@ -0,0 +1,65 @@
...
+     * This will create a new class instance base on an URLClassLoader using the <code>classPathUrls</code> and the
+     * <code>parentClassLoader</code>.
+     * This instance will be embed in a {@link SandboxedInstance} so that it provides a SystemProperty isolation.
+     * 
+     * @param classToInstanciated name of the class to instanciate
+     * @param classPathUrls set of URLs to be used for creating a new {@link ClassLoader}.
+     * @param parentClassLoader used a parent ClassLoader for the newly created classloader.
+     * @return a SandboxedInstance object ready for System Properties isolation
+     */
+    public static SandboxedInstance createSandboxedInstance(String classToInstanciated, Set<URL> classPathUrls,
```
RyanSkraba: URL order is relevant for an URLClasslaoder, so List or array would be better.  Maybe array since these are supposed to be precalculated anyway?

daikon/src/main/java/org/talend/daikon/sandbox/SandboxedInstance.java
```diff
@@ -0,0 +1,91 @@
...
+ * This object must be closed (see {@link #close()} so that the thread used when {@link #getInstance()} was called gets it's
+ * initial
+ * classloader back and that the isolation mechanism be disconnected from the classloader.
+ */
+public class SandboxedInstance implements AutoCloseable {
+
+    // this swith the current JVM System Properties with our own so that it can handle Thread/ClassLoader isolation
+    static {
+        ClassLoaderIsolatedSystemProperties isolatedSystemProperties = ClassLoaderIsolatedSystemProperties.getInstance();
+        if (System.getProperties() instanceof ClassLoaderIsolatedSystemProperties) {
```
RyanSkraba: Is this missing a NOT?

daikon/src/main/java/org/talend/daikon/sandbox/SandboxedInstance.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private Object instance;
+
+    ClassLoader previousContextClassLoader;
+
+    private ClassLoaderIsolatedSystemProperties isoSystemProp;
+
+    Thread isolatedThread;
+
+    SandboxedInstance(Object instance, ClassLoaderIsolatedSystemProperties isoSystemProp) {
```
RyanSkraba: The isoSystemProp is guaranteed to be the same singleton from the static block -- I'd get rid of it in the constructor and as a member.

daikon/src/main/java/org/talend/daikon/sandbox/properties/ClassLoaderIsolatedSystemProperties.java
```diff
@@ -0,0 +1,324 @@
...
+        if (instance == null) {
+            instance = new ClassLoaderIsolatedSystemProperties(System.getProperties());
+        }
+        return instance;
+    }
+
+    /**
+     * "Isolates" a Classloader: this means calls to this class methods will setup isolation mechanism so that
+     * classloader-specific properties be set with the <code>theClassLoaderProperties</code>.<br>
+     * WARNING : The isolation mechanism will only work <b>if and only if</b> the current Thread contextClassLoader is set to this
```
RyanSkraba: A very important point here -- this should be repeated in SandboxInstanceFactory.

daikon/src/main/java/org/talend/daikon/sandbox/SandboxedInstance.java
RyanSkraba: Actually, I'd put this in ClassLoaderIsolatedSystemProperties -- it should be clear that initializing the singleton there makes this global change for the current JVM process.

daikon/src/main/java/org/talend/daikon/sandbox/properties/ClassLoaderIsolatedSystemProperties.java
```diff
@@ -0,0 +1,324 @@
...
+            classLoaderProperties.put(classloader, theClassLoaderProperties);
+        }
+    }
+
+    /**
+     * disconnect the isolation mechanism from the classloader so that it will now work on shared properties.
+     *
+     * @param classLoader The {@link ClassLoader} to disconnect.
+     */
+    public void disconnectClassLoader(ClassLoader classLoader) {
```
RyanSkraba: Should be renamed (This does not disconnect the classloader!)   Maybe just startIsolate(CL, P) and stopIsolate(CL)?

daikon/src/main/java/org/talend/daikon/sandbox/SandboxedInstance.java
```diff
@@ -0,0 +1,91 @@
...
+     * Return the instance created by the {@link SandboxInstanceFactory} that now isolates the System Properties.
+     * <b>WARNING</b> : this also changes the current thread contextClassloader with the classloader used to create the instance,
+     * this
+     * will enable the isoltion to work. The contextClassLoader will be reset upon the class {@link #close()} call.<br>
+     * {@link #close()} must always be called.<br>
+     * <b>Please</b> read carefully the {@link #close()} javadoc.
+     * 
+     * @return the instance
+     */
+    public Object getInstance() {
```
RyanSkraba: This is a bit scary here -- getting the instance changes the state of the JVM, and this method can/MUST only be called exactly once (or the previousContextClassLoader is lost).  This is unexpected behaviour for an accessor!

The quick fix is to check whether the preparation (finding the previousContextClassLoader, etc) has already been performed.  You should also document that this means that the instance MUST be used only in one thread.

daikon/src/main/java/org/talend/daikon/sandbox/properties/StandardPropertiesStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+// along with this program; if not, write to Talend SA
+// 9 rue Pages 92150 Suresnes, France
+//
+// ============================================================================
+
+package org.talend.daikon.sandbox.properties;
+
+import java.util.Properties;
+
+public interface StandardPropertiesStrategy {
```
RyanSkraba: Do we really need a Strategy, StrategyFactory and an SunOracle implementation?

I suggest getting rid of these three classes in org.talend.daikon.sandbox.properties, then move the ClassLoaderIsolatedSystemProperties into org.talend.daikon.sandbox and make it package protected.

daikon/src/main/java/org/talend/daikon/sandbox/properties/SunOracleStandardPropertiesStrategy.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Properties;
+import java.util.StringTokenizer;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Implementation of {@link StandardPropertiesStrategy} for Sun/Oracle JVM (tested on 1.6)
+ */
+class SunOracleStandardPropertiesStrategy implements StandardPropertiesStrategy {
```
RyanSkraba: I'd prefer to use a copy of the existing System properties before classloading isolation takes place instead of trying to reparse the sun.conf file.  Wouldn't this implementation lose any parameters passed in via -Dmy.prop=myvalue?

fhuaulme:
Review done F2F, changes to be done by @sgandon

daikon/src/main/java/org/talend/daikon/sandbox/SandboxInstanceFactory.java
sgandon: Very good point, I'll fix this right away.

daikon/src/main/java/org/talend/daikon/sandbox/SandboxedInstance.java
sgandon: yes the not is missing and you are right I'll move it to ClassLoaderIsolatedSystemProperties

daikon/src/main/java/org/talend/daikon/sandbox/properties/ClassLoaderIsolatedSystemProperties.java
sgandon: I am not following you here ? can you elaborate please.

daikon/src/main/java/org/talend/daikon/sandbox/properties/ClassLoaderIsolatedSystemProperties.java
sgandon: yessss, thank you I was struggling with the name of this method just inherited from MDM implem.

daikon/src/main/java/org/talend/daikon/sandbox/SandboxedInstance.java
sgandon: yes, I'll do.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/83/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/83/)

RyanSkraba:
LGTM!

I have a small and **_optional_** suggestion -- the boolean useCurrentJvmProperties in SandboxInstance.getInstance(), can it be in the SandboxInstanceFactory.createSandboxedInstance() instead?  The two calls are *probably* made close together, so I'll leave it up to your preference!

sgandon:
I'll do it, I think it is better.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/85/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/85/)

PR #59, opened 2016-08-26
Feature/tdkn 73 tdkn 46 add configuration values for widget

ycbai:


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/91/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/91/)

daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
```diff
@@ -99,6 +105,15 @@
      **/
     public static final String ENUMERATION_WIDGET_TYPE = "widget.type.enumeration";
 
+    /*
+    Widget configurations
+     */
+
```
francisu: I don't understand why we need this configuration thing. Read only is the same idea as hidden for example. This will show up when this object is serialized using the normal mechanism. What's the additional purpose of the configuration values?

daikon/src/test/java/org/talend/daikon/properties/presentation/WidgetTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.talend.daikon.properties.Properties;
+import org.talend.daikon.properties.property.Property;
+import org.talend.daikon.properties.testproperties.TestProperties;
+
+import static org.junit.Assert.*;
+import static org.talend.daikon.properties.presentation.Widget.widget;
+import static org.talend.daikon.properties.property.PropertyFactory.newProperty;
+import static org.talend.daikon.properties.property.PropertyFactory.newString;
+
+/**
```
francisu: Please remove the "created by" comment. We don't use those.

daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
sgandon: @francisu some widgets need to be configured, for instance the TABLE_WIDGET automatically adds a button bar to add/remove items in the table. But some Component Developers need to have the bar removed so we need some way to configure the widgets this is the reason for this implementation.
Read_Only is not the same as Hidden ! hidden is not show to the user and read_only is show but not writable.
This will not be serialized anyway because we do not serialize form hence Widgets in Properties.toSerialized().

francisu: The Form and Widget are serialized for communication purposes with the web interface; they are not serialized when being stored. This is why we have the transient vs. permanent option in the serialization mechanism.

I agree that read only is not the same as hidden. My point is that the means of specifying hidden and read only are the same. If we have the is/setReadOnly, then we don't need the "configuration" stuff. I still don't see why we need this.  If you have other examples of needing to configure (like the table), then let's see the implementation of that, and I would be fine with a configuration mechanism that has widget type-specific options (like button bar, etc). I can also imagine that read only could be part of that (as it does not apply to all types of widgets, unlike hidden). However, I don't want both is/setReadOnly and the configuration stuff for the same thing.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/95/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/95/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/104/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/104/)

daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
sgandon: At first I wanted the READ_ONLY to be only one configuration property and then I suggested ycbai to make it also a method because it could be applied to all the widgets just like hidden and this would ease the developer life to have this method. 
Why do you feel this could not be applied to all widgets ?

PR #62, opened 2016-09-12
TUP-14173: add HIDE_TOOLBAR conf constant in Widget.

ycbai:


daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
```diff
@@ -150,6 +150,11 @@
     public static final String READ_ONLY = "widget.conf.readonly";
 
     /**
+     * Tell the client whether hide the toolbar or not.
```
sgandon: could you be more precise and explain on which component this have an impact.

daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
```diff
@@ -150,6 +150,11 @@
...
+     */
+    public static final String HIDE_TOOLBAR = "widget.conf.hide.toolbar";
```
sgandon: I think all widget configuration key should have a specific naming convention to no confuse them with widget names, something like HIDE_TOOLBAR_WIDGET_CONF.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/130/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/130/)

PR #69, opened 2016-10-17
fix (TDI-37525): add missing when DiOutgoingSchemaEnforcer transform from Avro-compatible to  Talend-compatible types

jzhao-talend:

- [x] The commit message follows Talend standard
- [x] Tests for the changes have been added (for bug fixes / features)

Because of BigDecimal and Character type save String type value in avro.
But in the DiOutgoingSchemaEnforcer  transform from Avro-compatible to  Talend-compatible types
We miss those 2 type convert.

**What is the new behavior?**

Add above missing

- [x] No

### DIFF
--- a/daikon/src/main/java/org/talend/daikon/di/DiOutgoingSchemaEnforcer.java
+++ b/daikon/src/main/java/org/talend/daikon/di/DiOutgoingSchemaEnforcer.java
@@ -12,7 +12,12 @@
 // ============================================================================
 package org.talend.daikon.di;
 
-import java.util.*;
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 import org.apache.avro.Schema;
 import org.apache.avro.Schema.Field;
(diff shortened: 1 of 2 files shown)


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/71/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/71/)

daikon/src/main/java/org/talend/daikon/di/DiOutgoingSchemaEnforcer.java
```diff
@@ -194,6 +199,10 @@ private Object transformValue(Object value, Field wrappedField, Field outField)
             return value instanceof Date ? value : new Date((Long) value);
         } else if ("id_Byte".equals(talendType)) { //$NON-NLS-1$
             return value instanceof Number ? ((Number) value).byteValue() : Byte.parseByte(String.valueOf(value));
+        } else if ("id_Character".equals(talendType) || "java.lang.Character".equals(javaClass)) {
+            return value instanceof Character ? value : ((String) value).charAt(0);
```
gonchar-ivan: Maybe, here is a proper place to use AvroConverter(s):
ConvertBigDecimal
ConvertCharacter (should be updated, because it is not properly implemented)

We can consider Talend types as DatumT and Avro types as AvroT

daikon/src/test/java/org/talend/daikon/di/DiOutgoingSchemaEnforcerTest.java
```diff
@@ -434,4 +436,62 @@ public void testValueConversion_toDate_javaClass() {
         assertThat(enforcer.get(0), instanceOf(Date.class));
         assertThat(enforcer.get(0), is((Object) currentDate));
     }
+
+    @Test
+    public void testValueConversionToDecimal() {
+        // The expected schema after enforcement.
```
gonchar-ivan: Schema is not changed during enforcement. So, this schema can't be "expected".

It is just "specification" schema (specified by user in Schema Editor)

Another thing: it is not talend6Schema. It is usual avro schema.
Talend6 means it should contain some DI specific things, but it doesn't contain

PR #70, opened 2016-10-18
TCOMP-313: Add "setVisible" method on Widget.

pbailly:
Add a couple of method into widget to improve the usability of the "refreshLayout" method.
The list of the added method is:
```
public Widget setHidden(Property<Boolean> hidden)
public Widget setHidden()
public Widget setVisible(boolean visible)
public Widget setVisible(Property<Boolean> visible)
public Widget setVisible()
```
It will change the code from:
```
if (useOtherConnection) {
    form.getWidget(useDataSource).setHidden(true);
    form.getWidget(dataSource).setHidden(true);
} else {
    form.getWidget(useDataSource).setHidden(false);
    form.getWidget(dataSource).setHidden(!useDataSource.getValue());
}

```
into:
```
if (useOtherConnection) {
    form.getWidget(useDataSource).setHidden();
    form.getWidget(dataSource).setHidden();
} else {
    form.getWidget(useDataSource).setVisible();
    form.getWidget(dataSource).setVisible(useDataSource);
}
```

This PR also contains small improvement in the Form class, adding a "setVisible" method and extending the "getWidget" method to any NamedThing.

### DIFF
--- a/daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
+++ b/daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
@@ -246,6 +246,15 @@ public class Form extends SimpleNamedThing implements ToStringIndent {
     }
 
     /**
+     * Change the visibility status of all of the {@code Form}'s widgets.
+     */
+    public void setVisible(boolean hidden) {
+        for (Widget w : getWidgets()) {
+            w.setVisible(hidden);
+        }
+    }
+
+    /**
      * Return widget name from its contents
      */
     private String getWidgetContentName(Widget widget) {
@@ -273,9 +282,9 @@ public class Form extends SimpleNamedThing implements ToStringIndent {
     }
 
     /**
-     * Return the {@link Widget} with the specified name.
+     * Return the {@link Widget} of a NamedThing.
      */
-    public Widget getWidget(Property<?> child) {
+    public Widget getWidget(NamedThing child) {
         return widgetMap.get(child.getName());
     }
 
--- a/daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
+++ b/daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
@@ -248,9 +248,9 @@ public class Form extends SimpleNamedThing implements ToStringIndent {
     /**
      * Change the visibility status of all of the {@code Form}'s widgets.
      */
-    public void setVisible(boolean hidden) {
+    public void setVisible(boolean visible) {
         for (Widget w : getWidgets()) {
-            w.setVisible(hidden);
+            w.setVisible(visible);
         }
     }
 
--- a/daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
+++ b/daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
@@ -289,6 +289,13 @@ public class Widget implements ToStringIndent {
         return hidden;
     }
 
+    /**
+     * return if the current Widget is visible or not.
+     */
+    public boolean isVisible() {
+        return !hidden;
+    }
+
     public String getWidgetType() {
         return widgetType;
     }
--- a/daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
+++ b/daikon/src/main/java/org/talend/daikon/properties/presentation/Widget.java
@@ -250,6 +250,41 @@ public class Widget implements ToStringIndent {
         return this;
     }
 
+    /**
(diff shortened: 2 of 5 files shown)


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/75/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/75/)

daikon/src/main/java/org/talend/daikon/properties/presentation/Form.java
```diff
@@ -246,6 +246,15 @@ public void setHidden(boolean hidden) {
     }
 
     /**
+     * Change the visibility status of all of the {@code Form}'s widgets.
+     */
+    public void setVisible(boolean hidden) {
```
sgandon: there is problem here, you have the parameter that is called *hidden*, this is counter intuitive, if you do
*setVisibile(true)*, hidden is true so this means you want to hide it or you said setVisible(true)

francisu:
I'm uncomfortable with the idea of two Beans that manipulate the same thing (hidden and visible). I think that adds confusion.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/77/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/77/)

pbailly:
Francis > The biggest usage of this hidden/visibility in the studio is to declare "if this checkbox is check, then I am visible" or "if this element is visible, I am visible too".
We need easy method express that, and the `setHidden(false)` or the `setHidden(!myBoolean.getValue())` are confusing.

The setHidden method is here to change the default visibility, which is also good, but will tend to increase the complexity on advanced code. We usually tend to have some condition like:
`SHOW_IF="(ADVANCED_SETTINGS_CHECK=='true' AND (isShow[SPARK_MODE]) AND (SPARK_MODE == 'CLUSTER' OR SPARK_MODE=='YARN_CLUSTER'))"`
Which can be translated to:
`setHidden(! (advancedSettingsCheck.getValue() && (!form.getWidget(sparkMode.getName()).isHidden()) && (sparkMode.getValue().equals("CLUSTER") || sparkMode.getValue().equals("YARN_CLUSTER"))))`
or without the not:
`setHidden( (!advancedSettingsCheck.getValue()) || (form.getWidget(sparkMode.getName()).isHidden()) || ((!sparkMode.getValue().equals("CLUSTER")) && (!sparkMode.getValue().equals("YARN_CLUSTER")))`

My solution allow us to remove the negation, which will tend to a "just a little bit more readable":
`setVisible(advancedSettingsCheck.getValue() && (!form.getWidget(sparkMode).isHidden()) && (sparkMode.getValue().equals("CLUSTER") || sparkMode.getValue().equals("YARN_CLUSTER")))`

(I may have to add the "isVisible()" method on widget...)

Another point is the current statistics of the method:
The "setHidden" method is used 47 times with the argument "true", 26 times with the argument "false", which make sense, since a property is "visible" by default.
But when you use a variable to set the current visibility, we have 22 occurrences of code where we want that the visibility of the property match another one and 10 occurrences of a code where we want to make the opposite.
I don't count here the complexes cases here, but as you can see, we need two times more the setVisible than the setHidden.

As a result, you can say that the "setVisible" add a confusion, but that is just a way to ease the code. However, if we keep just one of the two method, my personal preference would be to remove the "isHidden" to avoid double negative sentence.

francisu:
@pbailly I withdraw my objection. I think with TCOMP we have tried to craft the APIs so they are very compact and clear to use, and can reduce the amount of code that's required to code the component. What you are doing furthers this, even though it does add two bean properties that correspond to a single underlying field. I approve this change.

RyanSkraba:
I think this sounds fine...  This is a case where the extra "syntactic sugar" adds enough readability to offset the wider API!  Reading the code above is much easier, almost in english. 

The `setVisible(otherBooleanProperty)` is probably our number one use case!

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/79/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/79/)

pbailly:
I just added the method "isVisible" on Widget.
Like I said on my previous post, developers are often going to express "I want to show this field if this field another field is visible".
My new modification allow to transform the code from:
`form.getWidget(myValue1).setVisible(!form.getWidget(myValue2).isHidden());`
into:
`form.getWidget(myValue1).setVisible(form.getWidget(myValue2).isVisible());`

PR #103, opened 2016-11-26
Bugfix/6.3/tdi 37866 dynamic enforcer

gonchar-ivan:

- [x] The new code does not introduce new technical issues (sonar / eslint)

- [x] Bugfix
- [ ] Feature

There is an error in DynamicIndexMapperByName#computeIndexMap()

It computes indexMap incorrectly when design schema contains only dynamic field

**What is the new behavior?**

Fixed above bug

- [ ] Yes
- [x] No

**Other information**:

Related to https://jira.talendforge.org/browse/TDI-37866

### DIFF
--- a/daikon/src/main/java/org/talend/daikon/di/DynamicIndexMapperByName.java
+++ b/daikon/src/main/java/org/talend/daikon/di/DynamicIndexMapperByName.java
@@ -54,8 +54,9 @@ class DynamicIndexMapperByName implements DynamicIndexMapper {
         this.designSchema = designSchema;
         this.designFields = designSchema.getFields();
         this.designSchemaSize = designFields.size();
-        if (AvroUtils.isIncludeAllFields(designSchema)) {
-            dynamicFieldPosition = Integer.valueOf(designSchema.getProp(DiSchemaConstants.TALEND6_DYNAMIC_COLUMN_POSITION));
+        String dynamicFieldProperty = designSchema.getProp(DiSchemaConstants.TALEND6_DYNAMIC_COLUMN_POSITION);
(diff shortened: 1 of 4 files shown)


daikon/src/main/java/org/talend/daikon/di/DynamicIndexMapperByName.java
```diff
@@ -72,10 +72,8 @@
     public int[] computeIndexMap(Schema runtimeSchema) {
         int[] indexMap = new int[designSchemaSize + 1];
 
+        indexMap[dynamicFieldPosition] = DYNAMIC;
```
sgandon: and what about the *dynamicFieldPosition* to be out of bounds or never set ?
I think you could have an IndexOutOfBoundException here.
what do you think ?

gonchar-ivan: You are right.

Normally, _dynamicFieldPosition_ will be always set, because there is only one constructor and it should set it.
However, Constructor is not properly check whether _dynamicFieldPosition_ value is present in schema.

I'll fix it.

gonchar-ivan:
Fixed a case when _dynamicFieldPosition_ is not set. Code will throw
IllegalArgumentException("Design schema doesn't contain dynamic field")

As for out of bounds case:
I can't fix it inside IndexMapper code. IndexMapper doesn't know how to fix it.
 If incoming value is incorrect (out of bound) the only thing IndexMapper code could do is to throw Exception with proper meaning (Here, IndexOutOfBoundException ).
If value is incorrect it is a problem of client code, which sets that value.

So, IndexOutOfBoundException is normal situation here.

PR #124, opened 2017-01-09
feat(fileIO): cache SandboxedInstance.

pbailly:
Caching this object will allow the process to not compute it multiple time,
it will dramatically improve the computation time, since the class loading is
now done once by object type.

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)

Currently, a SandboxedInstance is created for each time there is a rest call. This create a delay of 1 seconds during the class loading and 2 second for the instantiation of a beam pipeline.

**What is the new behavior?**
Reusing existing SandboxedInstance allow us to highly improve this computation time (aka: transform it to 0 second)

- [x] Yes
- [ ] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: 
A `sandboxedInstance` will now still keep its `instance` element alive, even after a clean.

**Other information**:

### DIFF
--- a/daikon/src/main/java/org/talend/daikon/runtime/RuntimeUtil.java
+++ b/daikon/src/main/java/org/talend/daikon/runtime/RuntimeUtil.java
@@ -12,17 +12,25 @@
 // ============================================================================
 package org.talend.daikon.runtime;
 
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLStreamHandler;
+import java.net.URLStreamHandlerFactory;
+import java.util.HashMap;
+import java.util.Map;
+
 import org.ops4j.pax.url.mvn.Handler;
 import org.ops4j.pax.url.mvn.ServiceConstants;
 import org.talend.daikon.sandbox.SandboxInstanceFactory;
 import org.talend.daikon.sandbox.SandboxedInstance;
 
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.URLStreamHandler;
-import java.net.URLStreamHandlerFactory;
-
 public class RuntimeUtil {
+
+    /**
+     * TODO: This hashmap is currently containing a small amount of object.
+     * However, as the development is going to create more and more components, we need to add a purge mechnaism.
+     */
+    private static Map<RuntimeInfo, SandboxedInstance> classLoaderCache = new HashMap<>();
 
     static {
         // The mvn: protocol is always necessary for the methods in this class.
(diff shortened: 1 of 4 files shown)


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/121/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/121/)

sgandon:
Could you fill the template PR please ?

daikon/src/main/java/org/talend/daikon/sandbox/SandboxedInstance.java
```diff
@@ -89,7 +89,7 @@ public void close() {
      * @throws TalendRuntimeException is the class failed to be instanciated
      */
     public Object getInstance() {
-        if (isolatedThread == null) {
+        if (instance == null && isolatedThread == null) {
```
sgandon: There is pb here, the sandbox is base on the current thread to isolate the system properties.
So you must not use the same instance in different threads.
With you cache you allow for multiple threads to use the same instances.
There is also an issue with the close because once it is closed it is not sandboxed anymore.

daikon/src/main/java/org/talend/daikon/runtime/RuntimeUtil.java
```diff
@@ -12,18 +12,26 @@
...
+import org.talend.daikon.sandbox.SandboxInstanceFactory;
+import org.talend.daikon.sandbox.SandboxedInstance;
 
 public class RuntimeUtil {
 
+    /**
+     * TODO: This hashmap is currently containing a small amount of object.
+     * However, as the development is going to create more and more components, we need to add a purge mechnaism.
+     */
+    private static Map<RuntimeInfo, SandboxedInstance> classLoaderCache = new HashMap<>();
```
sgandon: You should be using a WeakHashMap to make sure that whenever some memory is required the reference object are collected.

pbailly: Sadly I cannot used the WeakHashMap. It will reset the map way too soon to be used here.
To be completely explicit, if I make two calls to the REST API with more than one second of interval, the data was erased.

sgandon: may be it depends on the way the JVM is configured for memory size and allocation ?

pbailly:
We need to cache only the sandboxedClassLoader.
New implementation available on https://github.com/Talend/daikon/pull/126

PR #132, opened 2017-01-19
Feature/TCOMP-335, make refs work for rest

sgandon:

- [x] Bugfix
- [ ] Feature

during the deserialization migration of delete fields for properties, the value passed was a JsonObject which is unusable by the Properties implementation cause it should not know about the serialization engine.

**What is the new behavior?**
Now there is an attempt to convert the JsonObject to a business object before calling the deletedField migration callback.

- [ ] Yes
- [x] No


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/203/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/203/)

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
```diff
@@ -94,7 +94,15 @@ public void fieldMissing(Object object, String fieldName, Object value) {
...
+            if (value instanceof JsonObject) {
+                // convert the JsonObject to a real object
+                deserialized = new com.cedarsoftware.util.io.JsonReader().jsonObjectsToJava((JsonObject) value);
+                if (deserialized instanceof JsonObject) {
+                    deserialized = null;
+                }
+            }
+            Boolean migrated = ((DeserializeDeletedFieldHandler) object).deletedField(fieldName, deserialized);
             if (migrated) {
                 migratedDeleted[0] = true;
```
pbailly: This is not realy due to this PR, but I have difficulties to understand why migratedDeleted is a table of 1 element.
Can't we change that to a simple boolean?

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
```diff
@@ -94,7 +94,15 @@ public void fieldMissing(Object object, String fieldName, Object value) {
...
-            Boolean migrated = ((DeserializeDeletedFieldHandler) object).deletedField(fieldName, value);
+            Object deserialized = value;
+            if (value instanceof JsonObject) {
+                // convert the JsonObject to a real object
+                deserialized = new com.cedarsoftware.util.io.JsonReader().jsonObjectsToJava((JsonObject) value);
+                if (deserialized instanceof JsonObject) {
+                    deserialized = null;
+                }
+            }
+            Boolean migrated = ((DeserializeDeletedFieldHandler) object).deletedField(fieldName, deserialized);
```
pbailly: Let's be pedantic: the boolean migrated is useless, you can directly do:
```migratedDeleted[0] = ((DeserializeDeletedFieldHandler) object).deletedField(fieldName, deserialized);```

File: daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java [hunk 3]
```diff
@@ -94,7 +94,15 @@ public void fieldMissing(Object object, String fieldName, Object value) {
             if (!DeserializeDeletedFieldHandler.class.isAssignableFrom(object.getClass())) {
                 return;
             }
-            Boolean migrated = ((DeserializeDeletedFieldHandler) object).deletedField(fieldName, value);
+            Object deserialized = value;
+            if (value instanceof JsonObject) {
+                // convert the JsonObject to a real object
+                deserialized = new com.cedarsoftware.util.io.JsonReader().jsonObjectsToJava((JsonObject) value);
+                if (deserialized instanceof JsonObject) {
+                    deserialized = null;
```
pbailly: I am not 100% sure, about that, but shouldn't be better to put a "return" here?

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
sgandon: nope, this is not the same, cause migratedDeleted[0] may be already true from another migration that happend before and would be set to false if migration did no happened which we don't want.

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
sgandon: this a flag used to maintain the migration state aver all the migrations calls so this must be a reference, I suppose we should have used some AtomicBoolean, this would have been more elegant.

daikon/src/main/java/org/talend/daikon/serialize/SerializerDeserializer.java
pbailly: Ok, but we still need a big refactoring of the "migratedDeleted" value, this will create a null pointer if the value is never updated.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/209/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/209/)

PR #237, opened 2017-05-16
fix(TDKN-158): make ValidationResult.OK deprecated this causing logic…

akhabali:
… issues

- [x] Bugfix
- [ ] Feature

https://jira.talendforge.org/browse/TDKN-158

**What is the new behavior?**
make ValidationResult.OK deprecated 
we provide a new method ok() to be used instead of ValidationResult.OK.
the ok() method deliver a new instance every time called to avoid the logic mismatch caused with the static shared instance ValidationResult.OK

**This ValidationResult.OK attribute should be removed in the incoming version**

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/daikon/src/main/java/org/talend/daikon/properties/ValidationResult.java
+++ b/daikon/src/main/java/org/talend/daikon/properties/ValidationResult.java
@@ -33,7 +33,17 @@ public class ValidationResult {
         ERROR
     }
 
+    /**
+     * <b>TDKN-158</b> - this causing validation error.<br/>
+     * As the OK instance is static it's can be modified while validation process and loose it's real sense
+     * Use {@link #ok()} method to get a new instance instead of that
+     */
+    @Deprecated
     public static ValidationResult OK = new ValidationResult().setStatus(Result.OK);
+
+    public static ValidationResult ok() {
+        return new ValidationResult().setStatus(Result.OK);
+    }
 
     @JsonIgnore
     public Result status = Result.OK;


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/58/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/58/)

daikon/src/main/java/org/talend/daikon/properties/ValidationResult.java
```diff
@@ -33,8 +33,18 @@
         ERROR
     }
 
+    /**
+     * <b>TDKN-158</b> - this causing validation error.<br/>
+     * As the OK instance is static it's can be modified while validation process and loose it's real sense
+     * Use {@link #ok()} method to get a new instance instead of that
+     */
+    @Deprecated
     public static ValidationResult OK = new ValidationResult().setStatus(Result.OK);
```
francisu: I would make this final, and I would change ValidationResult to not allow it to reset the status once it's been set. This will solve the problem and not break the API. In addition, you will not be creating a new object everytime you make an OK validation result, and it's possible this may be in some performance sensitive code.

akhabali: If we prevent ValidationResult from changing the status once it's been set. This means that we can't use the same instance to do farther validation and this will break the existing code logic.
In almost all components, we can find code like :
```
   ValidationResult validationResult = super.validate(runtimeContainer, properties);
   if (validationResult.getStatus() == ValidationResult.Result.ERROR) {
         return validationResult;
   }
   // do farther validation...
   // change the validationResult status & message if necessary and return the result
```

francisu: We could provide a subclass implementation for this case that would make the status immutable.

akhabali: We can do something like 
````
    public Result status = Result.OK;  //This already exist
    public final static ValidationResult OK = new ValidationResult() {
        @Override
        public ValidationResult setStatus(Result status) {
            if(this.status != null){
                return this;
            }
            return super.setStatus(status);
        }
    };
````
**But**
The problem is that the static `OK` attribute was mutable and it's used as a mutable instance in all the components. so changing it's behavior will break the logic in the component instead of breaking the daikon API.

For me not breaking the API means also keeping the same behavior (static mutable in this case), but this is causing bugs.

By the solution provided above in this comment, we have also to make the developers know about the behavior change of the `OK` attribute, so that they can refactor there code.

PR #244, opened 2017-05-29
feat(TCOMP-562): Store full information about properties errors and warnings.

dmytro-sylaiev:

- [ ] Bugfix
- [x] Feature

Properties.validationResult overwrites for every property

**What is the new behavior?**
ValidationResults class store all warnings and errors, user can update it or remove (if update error/warning to OK) 

- [ ] Yes
- [x] No

**Other information**:


daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,135 @@
...
+     * @return return list of all properties warnings (ValidationResult equals Result.Warning and message should be specified)
+     */
+    public List<ValidationResult> getAllWarningsAsList() {
+        return new ArrayList<>(warnings.values());
+    }
+
+    /**
+     * @return return list of all properties errors (ValidationResult equals Result.Error and error message should be specified)
+     */
+    public List<ValidationResult> getAllErrorsAsList() {
```
iistomin-gl: I would suggest to simplify name of the method, just getErrors(). There are no methods that return ERROR ValidationResults in some other form (not as List). The same suggestion applies to getAllWarningsAsList() method.

daikon/src/main/java/org/talend/daikon/properties/Properties.java
```diff
@@ -248,6 +248,9 @@
      */
     ValidationResult getValidationResult();
 
+
+    //TODO write javadoc
```
sgandon: please do so.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Contains and process validation result for every property
+ */
+public class ValidationResults {
+
+    private Map<String, ValidationResult> errors;
+    private Map<String, ValidationResult> warnings;
```
sgandon: why do you need 2 lists ? the class ValidationResult as already a status that represent the error level.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
sgandon: I agree

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,135 @@
...
+        if (warnings.containsKey(problemKey) || errors.containsKey(problemKey)) {
+            updateProblem(problemKey, validationResult);
+        }
+        else {
+            addProblem(problemKey, validationResult);
+        }
+        generalValidationResult = calculateValidationResult();
+    }
+
+    private void updateProblem(String problemKey, ValidationResult validationResult) {
```
sgandon: you don't need this method just the add method is enough

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,135 @@
...
+        errors.remove(problemKey);
+        warnings.remove(problemKey);
+
+        addProblem(problemKey, validationResult);
+    }
+
+    private void addProblem(String problemKey, ValidationResult validationResult) {
+        if (ValidationResult.Result.ERROR == validationResult.getStatus()) {
+            errors.put(problemKey, validationResult);
```
sgandon: this will update an existing key if it already exists, so you don't need the update method.

sgandon:
We really need to have a public access to the field names and their respective errors please add a public method in the ValidationResults to get access to those maps, you way even return an immutable map. 
but for the web client the Properties serialization will need to know which field has which error.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
dmytro-sylaiev: it is more convenient for client to have warnings and errors separately (i.e. in Studio now Node errors are shown first, Node warnings then) so if we store it like now, client (studio, webclient, etc) won't have to split it manually. 
I can unite it into 1 map but still leave List getWarnings() and List getErrors, but as I understand, after serialization all problems will store in a "pile"

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
dmytro-sylaiev: When we have 2 maps according to error and warning types, when update validation result from warning to error (or vice versa) we need also remove if from old collection before adding to new one.

sgandon: in fact I think Properties should not have a validation result field at all, they should be able to compute a ValidationResults at a given moment (from a method call) and never store this.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
sgandon: yes I prefer you merge them both and provide some help filtering those. This is a way better approach especially if you envision future changes like adding new statuses.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
sgandon: if you merge them both, there is no more issue :)

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -26,17 +23,26 @@ public ValidationResults() {
     /**
      * @return return list of all properties warnings (ValidationResult equals Result.Warning and message should be specified)
      */
-    public List<ValidationResult> getWarnings() {
+    public List<ValidationResult> getWarningsList() {
```
sgandon: why did you rename it ? *getWarnings* was fine to me.

daikon/src/main/java/org/talend/daikon/properties/Properties.java
```diff
@@ -248,6 +248,13 @@
      */
     ValidationResult getValidationResult();
 
+
+    /**
+     * Returns the {@link ValidationResults} for the component properties with errors and warnings when they occurs if added.
+     *
+     * @return a ValidationResult
+     */
+    ValidationResults getValidationResults();
```
sgandon: Why do we have to store this ?
I was wondering if it was more simpler to have a method  that computes on demand ValidationResults cause it does may eventually reflect the wrong status if one of the property has changed.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
dmytro-sylaiev: getWarnings() and getErrors() now return Map due to this requirements:
https://github.com/Talend/daikon/pull/244#issuecomment-304640714

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
dmytro-sylaiev: So do you advice to have private `Map<String, ValidationResult> problems` and have next getters:
`Map<String, ValidationResult> getProblems()` , `List<ValidationResult> getErrors()` , `List<ValidationResult> getWarnings()` ?

sgandon: yes

daikon/src/main/java/org/talend/daikon/properties/PropertiesImpl.java
```diff
@@ -511,6 +513,8 @@ public ValidationResult getValidationResult() {
         return validationResult;
     }
 
+    public ValidationResults getValidationResults() { return validationResults; }
```
gonchar-ivan: Please, use correct code formatter

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
+package org.talend.daikon.properties;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.*;
```
gonchar-ivan: we don't use * in imports. Please, extend concrete imports

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Contains and process validation result for every property
+ */
+public class ValidationResults {
+
+    private Map<String, ValidationResult> validationResults;
```
gonchar-ivan: We may have splitted collections (errors and warnings) of validation results in DTO if required. We need ask TCOMP service users which way is convenient for them: unite collection or splitted

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public ValidationResults() {
+        this.validationResults = new LinkedHashMap<>();
+    }
+
+    /**
+     * @return list of all properties warnings (ValidationResult equals Result.Warning and message should be specified)
+     */
+    public List<ValidationResult> getWarnings() {
+        List<ValidationResult> warnings = new ArrayList<>();
```
gonchar-ivan: I'm not sure what Java version is allowed to use in Daikon.
If java 8 is allowed we may use java streams feature to make this method shorter

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
gonchar-ivan: Please, add a javadoc to specify what should be used as **key** in this map.
Also maybe a **Property** could be used as a key?

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
...
+    public Map<String, ValidationResult> getAllValidationResults() {
+        return Collections.unmodifiableMap(validationResults);
+    }
+
+    /**
+     * Add problem to validationResults Map if it was absent there and status is WARNING or ERROR
+     * Update problem status and message if warning or error was present in Map
+     * Remove warning or error from map if new ValidationResult.status is OK
+     *
+     * @param problemKey not empty and not null
```
gonchar-ivan: Please, specify what String should be used as a key

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        if (null == validationResult) {
+            throw new IllegalArgumentException("Property Validation Result couldn't be null");
+        }
+
+        ValidationResult.Result vrStatus = validationResult.getStatus();
+
+        if ((ValidationResult.Result.ERROR == vrStatus || ValidationResult.Result.WARNING == vrStatus)
+                && StringUtils.isBlank(validationResult.getMessage())) {
```
gonchar-ivan: I suggest to split this check:

1. check status is ERROR or WARNING
2. check message isn't empty

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
...
+            throw new IllegalArgumentException("ValidationResult message couldn't be null or empty when status is " + vrStatus);
+        }
+
+
+        if (validationResults.containsKey(problemKey)) {
+            validationResults.remove(problemKey);
```
gonchar-ivan: Why do you need to remove?
I thought Map.put() will replace existing entry

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+         addProblem(problemKey, validationResult);
+
+    }
+
+    private void addProblem(String problemKey, ValidationResult validationResult) {
+        if (ValidationResult.Result.ERROR == validationResult.getStatus() || ValidationResult.Result.WARNING == validationResult.getStatus()) {
```
gonchar-ivan: IMO no need to check status again. This is a private method, so you know where it is called and you already checked status.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
...
+            validationResults.put(problemKey, validationResult);
+        }
+        //No need to store VR.OK in this class yet, only errors and warnings
+    }
+
+    /**
+     * @return actual ValidationResult depending on content of errors and warnings maps
+     */
+    public ValidationResult calculateValidationResult() {
```
gonchar-ivan: Do we need this feature at all? How will it be used?

Also there are several ways to implement it:
You add _status_ field and change it when add/remove error/warning validation result

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
...
+            status = ValidationResult.Result.WARNING;
+        }
+        return new ValidationResult(status, toString());
+    }
+
+    /**
+     *
+     * @return empty String when no problems in maps or full errors and warnings message
+     */
+    public String getGeneralProblemsMessage() {
```
gonchar-ivan: I'm not sure it should be public from start.
I have a rule to make public as less methods as I can. When you declare some method as public you must not remove/rename/change arguments of this method. So better to declare it less visibility unless you are sure it should be public

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,130 @@
+package org.talend.daikon.properties;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.*;
+
+/**
+ * Contains and process validation result for every property
+ */
+public class ValidationResults {
```
gonchar-ivan: What about to remove problems, when user have fixed Properties?

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
dmytro-sylaiev: If new validation result status is OK no need to store it anymore

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
dmytro-sylaiev: if message is empty for status OK it is not a problem

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
dmytro-sylaiev: We need to decide it :)

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
dmytro-sylaiev: `<java.version>1.7</java.version>`

sgandon: yes java 7

gonchar-ivan:
DO NOT MERGE BEFORE 0.19.0 RELEASE

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    private void handleOk(String problemKey) {
+        if (validationResults.containsKey(problemKey)) {
+            validationResults.remove(problemKey);
+        }
+    }
+
+    private void addProblem(String problemKey, ValidationResult validationResult) {
+        if (ValidationResult.Result.ERROR == validationResult.getStatus() || ValidationResult.Result.WARNING == validationResult
```
gonchar-ivan: This `if` is not required. We already know that it is error or warning.
See code line 77

daikon/src/test/java/org/talend/daikon/properties/ValidationResultsTest.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.List;
+
+/**
+ * Created by dmytro.sylaiev on 5/29/2017.
+ */
+public class ValidationResultsTest {
+
+    ValidationResults validationResults;
+
+    @Before public void setUp() {
```
gonchar-ivan: We used to put annotation in separate line:
```
@Before
public void setUp() {
```

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/252/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/252/)

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Contains and process validation result for every property
+ */
+public class ValidationResults {
+
+    /**
+     * Map containing validation results where key is the property name.
+     */
+    private Map<String, ValidationResult> validationResults;
```
gonchar-ivan: Consider to add `final`

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * @return map of all properties errors (ValidationResult equals Result.Error and error message should be
+     * specified)
+     */
+    public Map<String, ValidationResult> getErrors() {
+        return filterResultsByStatus(validationResults, Result.ERROR);
+    }
+
+    private Map<String, ValidationResult> filterResultsByStatus(Map<String, ValidationResult> results,
+            ValidationResult.Result status) {
```
gonchar-ivan: please, remove results parameter. `validationResults` field is accessible

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+        return results.entrySet().stream().filter(map -> map.getValue().getStatus().equals(status))
```
gonchar-ivan: please, rename `map` to `entry` or `pair`

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+     */
+    private boolean shouldRewrite(ValidationResult containing, ValidationResult newValue) {
+        return ((containing.getStatus() == Result.OK && newValue.getStatus() != Result.OK)
+                || (containing.getStatus() == Result.WARNING && newValue.getStatus() == Result.ERROR));
+    }
+
+    /**
+     * @return unmodifiable Map where keys are problemKeys and values are all ValidationResults
+     */
+    public Map<String, ValidationResult> getAllValidationResults() {
```
gonchar-ivan: I suggest you to rename it to `getAll()`
Thus, class will have following methods: getWarnings(), getErrors(), getAll()
Caller will have similar code:
```
    validationResults.getWarnings();
    validationResults.getAll();
```

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+        return results.entrySet().stream().filter(map -> map.getValue().getStatus().equals(status))
+                .collect(Collectors.toMap(map -> map.getKey(), map -> map.getValue()));
+    }
+
+    public void addValidationResults(ValidationResults value) {
+        if (value == null || value.isEmpty()) {
+            return;
+        }
+        Map<String, ValidationResult> allValidationResults = value.getAllValidationResults();
+        if (allValidationResults != null) {
```
gonchar-ivan: It is redundant check.
`value.getAllValidationResults()` will never return null.
I think developer should check for null only those values, which is not under his control (e.g. incoming parameters of public methods, but not private methods)

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+            for (Entry<String, ValidationResult> validationResult : allValidationResults.entrySet()) {
```
gonchar-ivan: maybe, stream can be used here too. But I don't know whether it is reasonable

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    public void addValidationResult(String propName, ValidationResult value) {
+        if (value == null) {
+            return;
+        }
+        ValidationResult containing = validationResults.get(propName);
+        if (containing == null || shouldRewrite(containing, value)) {
+            validationResults.put(propName, value);
+        }
```
gonchar-ivan: maybe, we need some logger here to notify, when result wasn't put in map. (I'm not sure. What do you think?)

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Check whether newer validation result status of processed property is more critical than the containing one. If so, we
+     * should update the value.
+     */
+    private boolean shouldRewrite(ValidationResult containing, ValidationResult newValue) {
```
gonchar-ivan: what if new result has the same criticity, but for another reason (another message)? 
Should we rewrite in this case?

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+     */
+    public ValidationResult.Result calculateValidationStatus() {
+        ValidationResult.Result status = ValidationResult.Result.OK;
+        if (!getErrors().isEmpty()) {
+            status = ValidationResult.Result.ERROR;
+        } else if (!getWarnings().isEmpty()) {
+            status = ValidationResult.Result.WARNING;
+        }
+        return status;
+    }
```
gonchar-ivan: You may get rid off local variable here:
```
       if (!getErrors().isEmpty()) {
            return ValidationResult.Result.ERROR;
        } else if (!getWarnings().isEmpty()) {
            return ValidationResult.Result.WARNING;
        }
        return ValidationResult.Result.OK;
```
Also you may extract methods:
`hasErrors()` and `hasWarnings()`
They may be useful for clients.
(you may make them package private for the moment. If some client will request we may methods public)

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
```diff
@@ -0,0 +1,126 @@
...
+    private String getGeneralProblemsMessage() {
+        StringBuilder message = new StringBuilder();
+
+        for (Object error : getErrors().values()) {
+            message.append(error.toString()).append("\n");
+        }
+        for (Object warning : getWarnings().values()) {
+            message.append(warning.toString()).append("\n");
+        }
+        return message.substring(0, message.length() - 1);
```
gonchar-ivan: What if there are no errors and warnings?
Do you have such test-case?
java.lang.StringIndexOutOfBoundsException may be thrown here
UPD. I saw it is private method.
Then I have a question is it possible map is not empty, but there are no errors and warnings? (several OK results are stored in map)

daikon/src/main/java/org/talend/daikon/properties/validation/ValidationHelper.java
```diff
@@ -0,0 +1,44 @@
...
+//
+// ============================================================================
+package org.talend.daikon.properties.validation;
+
+import org.talend.daikon.properties.AnyPropertyVisitor;
+import org.talend.daikon.properties.Properties;
+import org.talend.daikon.properties.ValidationResults;
+import org.talend.daikon.properties.property.Property;
+
+public class ValidationHelper {
```
gonchar-ivan: Utility classes not always good things.
What do you think about refactor this class into 2 classes:
ValidationStrategy interface and DefaultValidationStrategy (actually interface may be skipped as we can always add it later if required). 
Thus, we will be able to override this method. E.g. I can imagine some implementation which returns empty ValidationResults (skips validation).

If you decide to leave utility class as is, I suggest to make it final and add private empty constructor (as many static analyzers will suggest to do)

daikon/src/main/java/org/talend/daikon/properties/property/Property.java
```diff
@@ -497,6 +500,18 @@ public void encryptStoredValue(boolean encrypt) {
         // do nothing by default see StringProperty for an example
     }
 
+    public ValidationResult validate() {
```
gonchar-ivan: I think some javadoc is required here to explain developers how they should implement validation (explain about Validator)

daikon/src/main/java/org/talend/daikon/properties/property/Property.java
```diff
@@ -497,6 +500,18 @@ public void encryptStoredValue(boolean encrypt) {
...
+        if (validator == null) {
+            return ValidationResult.OK;
+        }
+        return validator.validate(getValue());
+    }
+
+    public Property<T> setValidator(Validator<T> validator) {
```
gonchar-ivan: Looks like a good idea! We may have common Property validators for generic purposes

daikon/src/main/java/org/talend/daikon/properties/validation/Validator.java
```diff
@@ -0,0 +1,21 @@
...
+// You should have received a copy of the agreement
+// along with this program; if not, write to Talend SA
+// 9 rue Pages 92150 Suresnes, France
+//
+// ============================================================================
+package org.talend.daikon.properties.validation;
+
+import org.talend.daikon.properties.ValidationResult;
+
+public interface Validator<T> {
```
gonchar-ivan: Please, add some javadoc for this interface to explain where it may be used

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/253/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/253/)

daikon/src/test/java/org/talend/daikon/properties/ValidationResultsTest.java
```diff
@@ -0,0 +1,159 @@
...
+        public ValidationResult afterNestedProperties() {
+            return ValidationResult.OK;
+        }
+
+        public ValidationResult validateNestedProperties() {
+            return nestedProperties.afterPropertyToValidate();
+        }
+
+        @Override
+        public ValidationResults validate() {
```
gonchar-ivan: It has no sense. If you remove the method, this class will have the same implementation of this method (inherited from parent)

daikon/src/test/java/org/talend/daikon/properties/ValidationResultsTest.java
```diff
@@ -0,0 +1,159 @@
...
+            return propertyToValidate.validate();
+        }
+
+    }
+
+    public static class EmptyStringValidator implements Validator<String> {
+
+        @Override
+        public ValidationResult validate(String value) {
+            ValidationResult result = ValidationResult.OK;
```
gonchar-ivan: you may get rid off local variable result here

daikon/src/test/java/org/talend/daikon/properties/ValidationResultsTest.java
```diff
@@ -0,0 +1,159 @@
...
+            if (value == null || value.isEmpty()) {
+                result = new ValidationResult(Result.ERROR, "Empty string");
+            }
+            return result;
+        }
+
+    }
+
+    @Test
+    public void testValidationResults() {
```
gonchar-ivan: Here you are testing several methods in one test-case: getAll(), getErrors(), getWarnings()
I think, ideally such test case should be splitted into several test-cases, but it's up to you as you it takes more time to write more smaller tests.
(when such small test-case fails you see where to look the problem. Test-case method name tells you, which method has an error. This is a benefit of unit-tests - quick error detection. Big test-case methods require a little bit more time to read test-case or stacktrace )

daikon/src/main/java/org/talend/daikon/properties/PropertiesImpl.java
```diff
@@ -757,4 +757,9 @@ boolean computeEqualityWith(Properties current, Properties other) {
 
     }
 
+    @Override
+    public ValidationResults validate() {
+        return new ValidationResults();
```
gonchar-ivan: Do we leave it as is?
Seems, ValidatonHelper was supposed to be used here?

chmyga: Validation helper is supposed to be used in the place, where validation is required to be called. ValidationHelper contains method, which passes AnyPropertyVisitor, which calls validate() methods for Properties and for Property instances. This is a default basic implementation and developers can override it to implement some complex Properties validation(for example use several fields to validate their values).

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
chmyga: In my opinion this implementation is a better way to split the code, as we may pass to this method not only validationResults field. We can pass any map containing validation results and filter it by status.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
chmyga: Yeah, I guess I've forgotten to remove it after all the changes. thanks!

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
chmyga: This won't make much sense as we won't see it anyway. I don't think user would be interested in the reason why some of the validation results weren't added.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
chmyga: I don't think we should. There are several problems related to this question which can lead us to idea of keeping list of validation results for every property. But IMO that would be unnecessary and too heavy-weight.

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
chmyga: Yeah, sorry, that wasn't my code. I'll check this situation and fix it.

daikon/src/main/java/org/talend/daikon/properties/property/Property.java
chmyga: Also, this way, we don't need to override Property class and we still can create validated Property in one line.

daikon/src/main/java/org/talend/daikon/properties/validation/ValidationHelper.java
chmyga: It's an interesting idea about making it not a utility class but a strategy. Though I'm not sure we will ever need it. It would be kind of over-engineering in my opinion. Seeing your later comment, I guess you think that this method will be called from Properties.validate() method, but that's not how it works. We will call ValidationHelper validateProperties() method which will then call validate() methods of Properties instances.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/256/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/256/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/257/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/257/)

daikon/src/main/java/org/talend/daikon/properties/ValidationResults.java
sgandon: There is not reason to have this parameter is the method is not public, if it stays private you don't need it.

daikon/src/main/java/org/talend/daikon/properties/validation/ValidationHelper.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Returns the {@link ValidationResults} for the component properties with errors and warnings if they occur.
+     *
```
sgandon: please add more javadoc to explain what this method does precisely.

daikon/src/test/java/org/talend/daikon/properties/ValidationResultsTest.java
sgandon: ?

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/265/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/265/)

daikon/src/test/java/org/talend/daikon/properties/ValidationResultsTest.java
gonchar-ivan: here, code may be changed in following way:
```
if (value == null || value.isEmpty()) {
    return new ValidationResult(Result.ERROR, "Empty string");
} else {
    return ValidationResult.OK;
}
```
Actually, I agree such change is not required here as method is rather simple.
Such change (refactoring) may be useful in some big and complicated methods. It won't give benefit itself, but it may allow further changes, which will help to simplify method.
So, it's ok to leave here as is.
(I just finished reading Martin Fowler's Refactoring book, so I'm at the stage of overusing these techniques))) )

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/266/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/266/)

PR #252, opened 2017-07-11
feat(TFD-1713): Create converter json to avro schema

zafkirTalend:

- [ ] The commit(s) message(s) follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md#commit-message-format) ?

- [ ] Bugfix
- [x] Feature

No converter json to avro schema.

**What is the new behavior?**

New converter json to avro schema.

- [ ] Yes
- [x] No

**Other information**:
https://jira.talendforge.org/browse/TFD-1695


build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/111/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/111/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/112/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/112/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/113/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/113/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/114/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/114/)

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
```diff
@@ -0,0 +1,212 @@
...
+        return fields;
+    }
+
+    /**
+     * Generate random.
+     *
+     * @param map to create random number
+     */
+    public String generateRandomNumber(Map.Entry<String, JsonNode> map) {
+        return (map.getKey() + "_" + new Random().nextInt(100));
```
pbailly: The name is using an identifier that may not be unique for big/complex avro structure. Could you change the "nextInt" into an UUID?

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/115/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/115/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/116/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/116/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/117/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/117/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/118/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/118/)

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
```diff
@@ -0,0 +1,196 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(AvroConverter.class);
+
+    private final ObjectMapper mapper;
+
+    /**
+     * Constructor
+     *
+     * @param mapper
+     */
+    public ConvertJson(final ObjectMapper mapper) {
```
gonchar-ivan: I suggest to add
```
public static createJsonSchemaInferrer() {
    return new JsonSchemaInferrer(new ObjectMapper());
}
```

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
```diff
@@ -0,0 +1,196 @@
...
+
+    @Override
+    public Class<String> getDatumClass() {
+        return null;
+    }
+
+    @Override
+    public String convertToDatum(Schema value) {
+        return null;
+    }
```
gonchar-ivan: I think we need use another interface.
Thus, these methods won't be required

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+        return fields;
+    }
+
+    /**
+     * Generate random.
+     *
+     * @param map to create random number
+     */
+    public String generateRandomNumber(Map.Entry<String, JsonNode> map) {
```
gonchar-ivan: I suggest to make this method package visible.
Thus, we tell for this class client that it is not API method (and we may refactor/rename/remove it easily).
But still we will be able to test it

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
```diff
@@ -0,0 +1,196 @@
...
+     *
+     * jsonNode parameter: {"a": {"b": "b1"}, "d": 100}
+     *
+     * jsonNode fields schema:
+     * [{"name":"a","type":{"type":"record","name":"a_49","fields":[{"name":"b","type":["null","string"]}]}},{"name":"d","type":["null","int"]}]
+     *
+     * @param jsonNode
+     * @return fields schema of json node
+     */
+    public ArrayNode getFields(final JsonNode jsonNode) {
```
gonchar-ivan: I suggest to make this method package visible.

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
```diff
@@ -0,0 +1,196 @@
...
+        return fields;
+    }
+
+    /**
+     * Generate random.
+     *
+     * @param map to create random number
+     */
+    public String generateRandomNumber(Map.Entry<String, JsonNode> map) {
+        return (map.getKey() + "_" + UUID.randomUUID().toString().replace("-", "_"));
```
gonchar-ivan: If it is possible and not hard to implement I suggest to construct field/schema name in following manner:
path + indexNumber
where path = outerRecord_level1Object_levelNObject_1
then 
outerRecord_level1Object_levelNObject_2
etc
Thus:
1. It will be unique
2. It will be clear where this object is located (path lists all his parents)
3. It will be easy for reading

gonchar-ivan: Also I suggest to rename method to something like:
String generateChildName()

Because, now it is strange that generateRandomNumber() method returns not Number, but String

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
```diff
@@ -0,0 +1,196 @@
...
+     * [{"name":"a","type":{"type":"record","name":"a_49","fields":[{"name":"b","type":["null","string"]}]}},{"name":"d","type":["null","int"]}]
+     *
+     * @param jsonNode
+     * @return fields schema of json node
+     */
+    public ArrayNode getFields(final JsonNode jsonNode) {
+        final ArrayNode fields = mapper.createArrayNode();
+        final Iterator<Map.Entry<String, JsonNode>> elements = jsonNode.fields();
+
+        Map.Entry<String, JsonNode> map;
```
gonchar-ivan: I suggest to rename it to `mapEntry`

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
gonchar-ivan: Also I would suggest to create Schema using Avro library classes and methods if it is possible.
Now, another schema JSON is created and then it is parsed to schema.
I think it is better to create Schema or Schema.Field right away.

Thus, you won't need ConvertJsonConstant.java class.

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
```diff
@@ -0,0 +1,196 @@
...
+        while (elements.hasNext()) {
+            map = elements.next();
+            final JsonNode nextNode = map.getValue();
+            ObjectNode fieldNode = mapper.createObjectNode();
+            ArrayNode fieldTypeArray = mapper.createArrayNode();
+
+            if (!(nextNode instanceof NullNode)) {
+                switch (nextNode.getNodeType()) {
+                case NUMBER:
+                    fieldNode.put(ConvertJsonConstant.NAME, map.getKey());
```
gonchar-ivan: I suggest to extract private methods from case branches

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJsonSchema.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.avro.generic.GenericRecordBuilder;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.NullNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.node.TextNode;
+
+public class ConvertJsonSchema {
```
gonchar-ivan: And this is JsonIndexedRecordConverter, which should implement AvroConverter<String, IndexedRecord>
where String is JSON

gonchar-ivan: Eventually, in future we may want to have a converter from IndexedRecord back to JSON representation.
So, I think this converter should be bidirectional

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
```diff
@@ -0,0 +1,212 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeType;
+import com.fasterxml.jackson.databind.node.NullNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+public class ConvertJson implements AvroConverter<String, Schema> {
```
pbailly: I really like your suggestion.

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
pbailly: I would also rename this method into something more relevant. Maybe "generateRecordName".

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
pbailly: I am against using the full tree to name a record, because we may want to reuse a subpart of an avro schema as its own schema.
I am in favor of the UUID (to be fair, I proposed it), this will prevent error when we are going to do complex merging operation like in the tMap.
Finally, I would argue that this name cannot be seen easily, so we can keep complex naming.

gonchar-ivan: yes, agree with you

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
sgandon: So here you say that there is not need to create a json to be serialized into a string to be converted into a Schema using Avro APIS but you would prefer the conversion to be done manually here ?

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
sgandon: please could you follow Ivan's suggestion.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/120/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/120/)

daikon/src/main/java/org/talend/daikon/avro/converter/ConvertJson.java
gonchar-ivan: As Is:
We construct JSON Object -> It is serialized to JSON String -> It is then parsed to Avro Schema
What I suggest:
We construct directly Avro Schema. Construction process is similar to constucting JSON Object

`final ArrayNode fields = mapper.createArrayNode();`
will become
`List<Schema.Field> fields = new ArrayList<>()`

```
case STRING:
               fieldNode.put(ConvertJsonConstant.NAME, map.getKey());
               fieldTypeArray.add(ConvertJsonConstant.NULL);
               fieldTypeArray.add(ConvertJsonConstant.STRING);
               fieldNode.put(ConvertJsonConstant.TYPE, fieldTypeArray);
               fields.add(fieldNode);
```
will become
```
case STRING: 
    field = new Schema.Field(map.getKey(), AvroUtils.wrapAsNullable(AvroUtils._string()), null, null, Order.ASCENDING);
```

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/121/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/121/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/122/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/122/)

daikon/src/main/java/org/talend/daikon/avro/converter/JsonGenericRecordConverter.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Override
+    public Schema getSchema() {
+        return null;
+    }
+
+    @Override
+    public Class<String> getDatumClass() {
+        return null;
```
sgandon: please return `String.class` here

daikon/src/main/java/org/talend/daikon/avro/converter/JsonGenericRecordConverter.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Override
+    public String convertToDatum(GenericRecord value) {
+        return null;
```
sgandon: here you could convert the record into json too.

daikon/src/main/java/org/talend/daikon/avro/converter/JsonGenericRecordConverter.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Override
+    public GenericRecord convertToAvro(String json) {
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            JsonNode jsonNode = mapper.readTree(json);
+            return getOutputRecord(jsonNode, schema);
+        } catch (Exception ex) {
```
sgandon: please never catch Exception, catch the appropriate exception here if any.
see page 12 for explanation on why this is a bad habit : https://wiki.talend.com/download/attachments/1643991/Java_Best_practice.pdf

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/123/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/123/)

daikon/src/main/java/org/talend/daikon/avro/inferrer/JsonSchemaInferrer.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * @return new JsonSchemaInferrer instance
+     */
+    public static JsonSchemaInferrer createJsonSchemaInferrer() {
+        return new JsonSchemaInferrer(new ObjectMapper());
+    }
+
+    /**
+     * Convert json string to avro schema.
```
pbailly: Could you actualise this comments?
Maybe: "Create an Avro Schema based on a JSON String."

daikon/src/main/java/org/talend/daikon/avro/converter/JsonGenericRecordConverter.java
```diff
@@ -51,26 +62,32 @@ public JsonGenericRecordConverter(Schema schema) {
...
+     */
     @Override
     public GenericRecord convertToAvro(String json) {
         try {
             ObjectMapper mapper = new ObjectMapper();
             JsonNode jsonNode = mapper.readTree(json);
             return getOutputRecord(jsonNode, schema);
-        } catch (Exception ex) {
+        } catch (IOException ex) {
             logger.error(ex.getMessage());
```
sgandon: actually I wonder if this is the best to handle this exception.
Ok the contract of this API is not very clear in the interface definition, but I am not sure a null value is expected when this API is used.
Especially for low level API like this I wonder if we should not throw a TalendRuntimeException that nests the IOException so that the exception is handled by the caller. This also saves from a null check.
And when something gets wrong you don't want to return null.
@gonchar-ivan what do you think ?

sgandon:
I have approved the changes but there is still one point that bothers me, see my last comment.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/124/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/124/)

daikon/src/main/java/org/talend/daikon/avro/converter/JsonGenericRecordConverter.java
gonchar-ivan: Yes, null is not expected result. It will be hard to find where problem appeared if we return null here.
As I can imagine ObjectMapper throws IOException in case of wrong JSON.

Better to throw exception. Let it be TalendRuntimeException for the moment... However I think it is also not the best solution as usually it is showed as UNEXPECTED_EXCEPTION without any information.
I think we need to set proper error_codes and context for the Exception.
Eventually I'm thinking about extending TalendRuntimeException.
We may have something like DataConvertException. But this solution is not ready for the moment, so let's use TalendRuntimeException

sgandon: ok this will require also a change in the Converter interface javadoc to say it never returns null but throws a TalendRuntimeException.

sgandon:
@zafkirTalend can you have a look at those comments : https://github.com/Talend/daikon/pull/252#discussion_r127236961
they are hidden by github cause the code changed but could you take them into account and this PR should be fine. thanks.

daikon/src/main/java/org/talend/daikon/avro/converter/JsonGenericRecordConverter.java
zafkirTalend: Ok thanks.

zafkirTalend: I'll do that after extending TalendRuntimeException.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/125/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/125/)

sgandon:
thanks a lot Zakaria.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/126/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/126/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/127/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/127/)

PR #262, opened 2017-07-31
when there is a single form and schemagenerator gets a properties use that form

rmannibucau:

- [ ] The commit(s) message(s) follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md#commit-message-format) ?

- [X] Bugfix
- [ ] Feature

When a properties is the content of a widget the form can't be browsed in a complex form structure and the generation fails with a NPE.

**What is the new behavior?**

If the properties has a single form it is automatically used otherwise we fail explicitly.

- [ ] Yes
- [X] No

-

**Other information**:

This is a proposal, alternative is to say salesforce component was wrongly written and needs to be fixed. This also throw the question of the layout model which seems to mix different kind of models leading to this kind of issue with complex UI. Can need to revise the API to prevent it.


build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/159/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/159/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/164/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/164/)

daikon/src/main/java/org/talend/daikon/serialize/jsonschema/JsonSchemaGenerator.java
```diff
@@ -111,8 +111,17 @@ private ObjectNode processTProperties(Properties cProperties, String formName, b
...
+                if (isVisible) {
+                    if (Form.class.isInstance(widget.getContent())) {
+                        propertiesFormName = widget.getContent().getName();
+                    } else if (Properties.class.isInstance(widget.getContent())) { // likely properties
+                        final Properties props = Properties.class.cast(widget.getContent());
+                        if (props.getForms().size() == 1) {
+                            propertiesFormName = props.getForms().iterator().next().getName();
+                        } else {
+                            throw new IllegalArgumentException("You need to wrap " + properties + " in a form in " + widget);
+                        }
```
francisu: Are you sure the properties.toString() and widget.toString() is meaningful with this exception (there should be a test for this exception case)? I don't see that test. Also, we don't generally say things like "You need to" in an exception message, it could say what it found and suggest what to do. For example, something like: "The Widget: " widget " directly references Properties: " properties " but there is not a single Form associated with the properties object. Please specify a Form object for the Widget."

rmannibucau:
There is an issue with this patch since it changes names of properties. As mentionned in the _other information_ comment, it probably requires to rethink the model more deeply either on the form/widget area or the  ui/json area. This PR contains a test reproducing the issue which can be reused if needed but will close it as it can't be solved cleanly directly without breaking existing internationalization not relying on this defaulting form logic.

PR #282, opened 2017-09-22
feat(avro): provide a visitor pattern for Avro generic records

bguillon-td:

It's really cumbersome to handle hierarchical Avro Generic records (IndexedRecord) using the native Avro API as it does not provide a nice way to visit a generic record.

With this pull request, it is now possible to visit a Avro generic record using the following API
```java
IndexedRecord yourRecord = ...;
RecordVisitor yourVisitor = ...;
new VisitableRecord(yourRecord).accept(yourVisitor);
```
All implementation is in org.talend.avro.visitor.

This pull request also provides org.talend.avro.path which is an implementation of a generic traversal path API for generic records visit (*not a path expression*).
 
**Link to the JIRA issue**

 

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/261/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/261/)

bguillon-td:
No issue found by Codacy - but still unhappy.

RyanSkraba:
Hello!  Just some comments about this PR and how it relates to expression languages!

Pretend that avpath was selected as _the_ chosen expression language tech of choice by the tab task force.  In this case, this PR would have the following problems:

* The examples for TraversalPath -> JSONPath or XPath are tempting alternatives *against* using avpath expression language for Talend products.

* I don't know if this is a goal of this code, but if you're visiting a record to provide a list of possible XPath (or other) in the UI, you should probably be getting this info based on a Schema as opposed to a Record
  - make results reusable between records (i.e. detect paths once on schema, reuse on each row).
  - with the current code, records with the same schema will have different detected paths.  Is this a useful feature?  I can see it being used for column headings, but data-streams is already hitting some limits with this approach.

* (Important) Developers really shouldn't be tempted to use the visitor pattern as a "quick and dirty expression parser".  
  - i.e., given an XPath expression, visit the record until you find the TraversalPath that evaluates to that expression and return the value.
  - I can give details about how we do something _almost_ as dubious in the Studio with tAvroMap (but with some improvements).

* (Just an opinion) providing an abstraction layer over the choice of expression language is probably architectural overkill.

None of these issues are related to the usability of the feature for visiting records.

sgandon:
In fact I am wondering what is the use of the TraversalPath as a public API ?
I wonder if you should not remove the *getPath* from `org.talend.daikon.avro.visitor.VisitableStructure<T>` and make the `org.talend.daikon.avro.visitor.AbstractVisitableStructure.getPath()` package private.

bguillon-td:
Hi,
Thanks for your comments.

I've made the following changes in the package layout:
- moved the path implementation into org.talend.daikon.avro.visitor.path, it should show (but not enforce) that this path framework is limited to visitors implementation.
- moved all records visitor implementation into org.talend.daikon.avro.visitor.record so that we will have room to implement a Schema visitor as part of this framework.

As discussed with you @sgandon, retrieving the full visit path of a node during the record's visit is convenient for visitors implementations. I think that removing the path from the API would lead implementations to also implement their own path building strategies and so potential duplicate code or duplicate efforts.

PR #286, opened 2017-10-02
Igonchar/tcomp 611 extend api with runtime container arg

gonchar-ivan:

There is no way to pass some information from Product to Properties instance, but it is required for JDBC components to get location of user configuration.
 
This PR introduces RuntimeContext. It is container for key-value pairs, which Product may pass to Properties. Also overloaded callback methods were added to PropertiesService to allow passing RuntimeContext argument to Properties.

**Link to the JIRA issue**

https://jira.talendforge.org/browse/TCOMP-611
 
It should not introduce breaking change.


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/268/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/268/)

daikon/src/main/java/org/talend/daikon/properties/runtime/RuntimeContext.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    Object getData(String key);
+
+    /**
+     * Sets a data value for given key
+     * 
+     * @param key data value key
+     * @param value data value
+     */
+    void setData(String key, Object value);
```
sgandon: I am not sure why components developers would need such a *setData* ?
I would imagine the context is something ephemeral during the time of the call and is set by the components container and shall not be tempered by the components. Especially in multi-instances where the next request may not be made on this service instance.

sgandon:
I like the idea cause it is simple and does not bring any breaking changes.
See my comment though.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/270/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/270/)

daikon/src/main/java/org/talend/daikon/properties/runtime/RuntimeContext.java
gonchar-ivan: Yes, you are right. Component developers don't need such method. Only components container requires some means to setup the context. 
It can be done in following ways:
1. Inner anonymous interface implementation, in which getData() method returns required data
2. Setter methods may be implemented in specific implementations.

I'll remove setData() method from interface

daikon/src/main/java/org/talend/daikon/properties/PropertiesDynamicMethodHelper.java
```diff
@@ -20,33 +20,64 @@
...
+     * @param methodType method type, i.e. prefix used before property name part in method name
+     * @param propertyName name of property, which owns the trigger
+     * @param required specify whether method is required
+     * @param the list of parameter
+     * @return found method or <code>null</code> if method wasn't found and not required
+     */
+    static Method findMethod(Object instance, String methodType, String propertyName, boolean required,
+            Class<?>... parameterTypes) {
+        if (instance == null) {
+            throw new NullPointerException("Instance whose method is being searched for should not be null");
```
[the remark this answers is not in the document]
gonchar-ivan: This PMD rule is not fully correct or well explained.
Here, NPE is thrown with message. Thus, user won't confuse it with NPE thrown by JVM.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/274/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/274/)

daikon/src/main/java/org/talend/daikon/properties/runtime/RuntimeContextImpl.java
```diff
@@ -0,0 +1,40 @@
...
+// ============================================================================
+package org.talend.daikon.properties.runtime;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * {@link RuntimeContext} default implementation based on {@link HashMap}
+ */
+public class RuntimeContextImpl implements RuntimeContext {
```
gonchar-ivan: @sgandon 
What do you think about this common implementation of interface?
Is it required?

I have following thoughts about this:

1. If product pass only 1-2 values, then it may implement the interface as inner anonymous class:
```
String configLocation = ... // compute config location
componentService.afterProperty("prop", props, new RuntimeContext {
    @Override
    public Object getData(String key) {
        if (key.equals(CONFIG_LOCATION)) {
            return configLocation;
        }
    }
});
```

2. This RuntimeContext may be used for some advanced features. E.g. to store connection object or to cache some  data.
There was a case with Netsuite component. Similarly to Salesforce, there are custom modules (custom data schemas).
There is design time button, which allows to get all available modules (both pre-installed and custom) and choose one. It takes a lot of time (from 1 to 3 minutes) to get module names and schemas, because of slow processing from Netsuite side. So, it was reasonable to cache results retrieved after first call, but we have no such posibility.
RuntimeContext, may allow component to cache such data (then setData() method is required). In this case Product will need to store RuntimeContext instance for some time. Here, Product may require some common implementation to minimize its efforts. 
(Note, I am not sure storing component data in RuntimeConvext is not good idea as components/components service are stateless. This approach is applicable only for Studio)

I'm more inclined to remove common implementation for the moment.

daikon/src/test/java/org/talend/daikon/properties/PropertiesDynamicMethodHelperTest.java
```diff
@@ -0,0 +1,1356 @@
...
+        verify(withBothCallbacks).beforeProperty(context);
+    }
+
+    /**
+     * Asserts that method throws {@link IllegalArgumentException}, when {@link Properties} class has no triggers at all
+     * 
+     * @throws Throwable
+     */
+    @Test
+    public void testValidatePropertyNoCallback() throws Throwable {
```
gonchar-ivan: Client code should call isCall*() methods of Form and Widget classes to discover whether it is allowed to call callbacks.

There are 4 cases for each callback:
1. Properties implementation has no callbacks. isCall() should return false. Exception will be thrown if client call callback
2. Properties implementation has callback without RuntimeContext parameter (old callback). This is current implementation. It should work as before. isCall() should return true. Old callback may be called.
3. Properties implementation has callback with RuntimeContext parameter (new callback), but has no old callback. It breaks current implementation as product may call service without passing RuntimeContext. So, it should be prohibited. isCall() should return false. Exception should be thrown, when client calls the callback.
4. Properties has both callbacks. This correct implementation to support new feature. isCall() should return true. Product may pass RuntimeContext argument. Then new callback will be called. If Product doesn't pass RuntimeContext argument, then old callback should be called.

I'll add this description as type javadoc for this test class.
Also, seems there is a sense to modify test cases to check isCall() methods

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/275/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/275/)

sgandon:
I am not sure this PR will resolve anything, today we can use system properties and config files in the classpath. Here you are just adding more complexity on the container of the service ad I don't really see the gain.
could you elaborate ?

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/276/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/276/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/277/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/277/)

gonchar-ivan:
From discussion with Seb:
User specific settings is a good reason for keeping this because, despite you can change the System Properties on the fly but this is usually not recommended. Further more, today the component use the SandboxInstance that isolates the System properties.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/292/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/292/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/311/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/311/)

PR #324, opened 2017-12-18
fix(tenancy): use optional tenant to avoid NPE at runtime

afourmi:
 
**Link to the JIRA issue**

 
- [ ] The PR commit message follows our [guidelines](../CONTRIBUTING.md)

 

daikon-multitenant/multitenant-core/src/main/java/org/talend/daikon/multitenant/datasource/TenantAwareDataSource.java
```diff
@@ -28,7 +28,7 @@
     protected String getDatabaseName() {
         String dbName = null;
         TenancyContext tenancyContext = TenancyContextHolder.getContext();
-        if (tenancyContext.getTenant() != null) {
+        if (tenancyContext.getOptionalTenant().isPresent()) {
```
gonzalad: if tenant is mandatory when calling getDatabaseName() (afaik it is), best is to keep calling 
```
tenancyContext.getTenant()
```

This way you have an immediate and understandable error. Otherwise, you'll get a NPE downstream which will be difficult to diagnose.

PR #333, opened 2018-01-11
A filter/valve which puts some of the http headers into slf4j MDC

IevgenTarasov:
We need to automatically collect some of the data that comes with incoming http request (like user agent and client ip address). With filter/valve it will be done automatically.

**Link to the JIRA issue**
https://jira.talendforge.org/browse/TPSVC-3900
 
### DIFF
--- a/daikon-logging/logging-http-headers/README.adoc
+++ b/daikon-logging/logging-http-headers/README.adoc
@@ -2,7 +2,7 @@
 
 This library provides a servlet filter and a tomcat valve which populate slf4j MDC with data from HTTP request.
 Both filter and valve have the same functionality, and provided together for better flexibility (some apps will prefer
-the filter, while the other will prefer the valve).
+the filter, while the other will prefer the valve). Generally speaking, filter should be preferred over valve.
 
 The next fields are populated:
 
@@ -56,14 +56,10 @@ NOTE: This approach is not recommended, as there may be class-loading issues wit
            replaceRemoteAddrWithForwardedFor="false" />
 ----
 
-=== Spring Boot
+=== Filter with Spring Boot
 
-With spring boot it's possible to register filter or valve programmatically. For this it's necessary
-to define corresponding bean (see below) in a spring configuration class.
-
-==== Filter
-
-To define the filter dynamically:
+With spring boot it's possible to register filter programmatically. For this it's necessary
+to define corresponding bean (see below) in a spring configuration class:
 
 [source,java]
 ----
@@ -78,27 +74,6 @@ To define the filter dynamically:
         regBean.setOrder(SecurityProperties.DEFAULT_FILTER_ORDER + 1);
 
         return regBean;
-    }
-----
-
-==== Valve
-
-To define the valve dynamically:
-
-[source,java]
-----
-    @Bean
-    public EmbeddedServletContainerCustomizer tomcatContextCustomizer() {
-        HttpHeadersMDCValve valv = new HttpHeadersMDCValve();
-        valv.setReplaceRemoteAddrWithForwardedFor(false);
-
-        return (x -> {
-            if (x instanceof TomcatEmbeddedServletContainerFactory) {
-                ((TomcatEmbeddedServletContainerFactory) x).addContextCustomizers(y -> {
-                    y.getPipeline().addValve(valv);
-                });
-            }
-        });
     }
 ----
 
--- a/daikon-logging/logging-http-headers/README.adoc
+++ b/daikon-logging/logging-http-headers/README.adoc
@@ -1,0 +1,114 @@
+= Logging HTTP Headers
+
+This library provides a servlet filter and a tomcat valve which populate slf4j MDC with data from HTTP request.
+Both filter and valve have the same functionality, and provided together for better flexibility (some apps will prefer
+the filter, while the other will prefer the valve).
+
+The next fields are populated:
+
+|===
+|Name|MDC|Comment
+
+|Remote IP address|httpRemoteAddr|
+|Remote port|httpRemotePort|
+|User agent|httpUserAgent|From User-Agent header
+|Authenticated user|httpRemoteUser|The user associated with current http request, if any
+|X-Forwarded-For header|httpForwardedFor|This header is usually set by proxies when relay a request
+|===
+
+== Usage
+
+=== Static
+
+To use the lib statically it needs to be added to a config file either in web app or in tomcat.
+
+==== Filter
+
+The filter needs to be defined in the web.xml file of the web app that wishes to use it. For example:
+
+[source,xml]
+----
+    <filter>
+        <filter-name>httpHeadersMDCFilter</filter-name>
+        <filter-class>org.talend.daikon.logging.http.headers.HttpHeadersMDCFilter</filter-class>
+        <init-param>
+            <param-name>replaceRemoteAddrWithForwardedFor</param-name>
+            <param-value>false</param-value>
+        </init-param>
+    </filter>
+    <filter-mapping>
+        <filter-name>httpHeadersMDCFilter</filter-name>
+        <url-pattern>/*</url-pattern>
+    </filter-mapping>
+----
+
+==== Valve
+
+The valve needs to be defined in tomcat's `conf/server.xml` file, under Engine section (it will be applied to
+all web apps deployed on this tomcat deployment). Additionally, the jar file with the valve (and possibly
+other dependencies, like `slf4j-api`) needs to be copied into `tomcat/lib` directory.
+
+NOTE: This approach is not recommended, as there may be class-loading issues with slf4j.
+
+[source,xml]
+----
+    <Valve className="org.talend.daikon.logging.http.headers.HttpHeadersMDCValve"
+           replaceRemoteAddrWithForwardedFor="false" />
+----
+
+=== Spring Boot
+
+With spring boot it's possible to register filter or valve programmatically. For this it's necessary
+to define corresponding bean (see below) in a spring configuration class.
+
+==== Filter
+
+To define the filter dynamically:
+
+[source,java]
+----
+    @Bean
+    public FilterRegistrationBean filterRegistrationBean() {
+        FilterRegistrationBean regBean = new FilterRegistrationBean();
+
+        HttpHeadersMDCFilter filtr = new HttpHeadersMDCFilter();
+        filtr.setReplaceRemoteAddrWithForwardedFor(false);
+
+        regBean.setFilter(filtr);
+        regBean.setOrder(SecurityProperties.DEFAULT_FILTER_ORDER + 1);
+
+        return regBean;
+    }
+----
+
+==== Valve
+
+To define the valve dynamically:
+
+[source,java]
+----
+    @Bean
+    public EmbeddedServletContainerCustomizer tomcatContextCustomizer() {
+        HttpHeadersMDCValve valv = new HttpHeadersMDCValve();
+        valv.setReplaceRemoteAddrWithForwardedFor(false);
+
+        return (x -> {
+            if (x instanceof TomcatEmbeddedServletContainerFactory) {
+                ((TomcatEmbeddedServletContainerFactory) x).addContextCustomizers(y -> {
+                    y.getPipeline().addValve(valv);
+                });
+            }
+        });
+    }
+----
+
+=== Configuration
+
+The next table lists available configuration properties.
+
+|===
+|Property|Filter/Valve|Optionial?|Description
+
+|replaceRemoteAddrWithForwardedFor|Both|Yes (default: false)|Substitutes remote IP address with the value
+in X-Forwarded-For header (if available)
+|===
--- a/daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDC.java
+++ b/daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDC.java
@@ -1,0 +1,20 @@
+package org.talend.daikon.logging.http.headers;
+
+/**
+ *
+ */
+public final class HttpHeadersMDC {
+
+    public static final String HTTP_REMOTE_USER = "httpRemoteUser";
+
+    public static final String HTTP_REMOTE_ADDR = "httpRemoteAddr";
+
+    public static final String HTTP_REMOTE_PORT = "httpRemotePort";
+
+    public static final String HTTP_USER_AGENT = "httpUserAgent";
+
+    public static final String HTTP_FORWARDED_FOR = "httpForwardedFor";
+
+    private HttpHeadersMDC() {
+    }
+}
--- a/daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilter.java
+++ b/daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilter.java
@@ -1,11 +1,12 @@
 package org.talend.daikon.logging.http.headers;
 
-import org.slf4j.Logger;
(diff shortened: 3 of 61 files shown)


build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/92/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/92/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/93/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/93/)

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDC.java
```diff
@@ -0,0 +1,20 @@
+package org.talend.daikon.logging.http.headers;
+
+/**
+ *
+ */
+public final class HttpHeadersMDC {
```
fhuaulme: There's a class named `org.talend.daikon.logging.event.field.MdcKeys`, shouldn't these constants be moved to that class?

daikon-logging/logging-http-headers/README.adoc
```diff
@@ -0,0 +1,114 @@
...
+With spring boot it's possible to register filter or valve programmatically. For this it's necessary
+to define corresponding bean (see below) in a spring configuration class.
+
+==== Filter
+
+To define the filter dynamically:
+
+[source,java]
+----
+    @Bean
```
fhuaulme: This code snippet could come in a @Configuration-annotated class to prevent code duplication across projects.

And by doing so, other projects will only need to add the module to their classpath and it will automagically work.

daikon-logging/logging-http-headers/README.adoc
```diff
@@ -0,0 +1,114 @@
...
+        regBean.setFilter(filtr);
+        regBean.setOrder(SecurityProperties.DEFAULT_FILTER_ORDER + 1);
+
+        return regBean;
+    }
+----
+
+==== Valve
+
+To define the valve dynamically:
```
fhuaulme: It would be nice to list pros for this approach (I don't see what additional value it brings compared to the servlet Filter approach).
And moreover, example is tied to Tomcat.

daikon-logging/logging-http-headers/README.adoc
```diff
@@ -0,0 +1,114 @@
...
+    </filter>
+    <filter-mapping>
+        <filter-name>httpHeadersMDCFilter</filter-name>
+        <url-pattern>/*</url-pattern>
+    </filter-mapping>
+----
+
+==== Valve
+
+The valve needs to be defined in tomcat's `conf/server.xml` file, under Engine section (it will be applied to
```
fhuaulme: I did a comment about the pros/cons of this approach, I see only cons (you stated this approach is "not recommended", so why writing about it?)

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilter.java
```diff
@@ -0,0 +1,57 @@
...
+    public void setReplaceRemoteAddrWithForwardedFor(boolean replaceRemoteAddrWithForwardedFor) {
+        this.replaceRemoteAddrWithForwardedFor = replaceRemoteAddrWithForwardedFor;
+        LOG.debug("Setting {} to {} via setter", REPLACE_PARAM_NAME, replaceRemoteAddrWithForwardedFor);
+    }
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+        LOG.debug("Initializing filter {}", filterConfig.getFilterName());
+
+        final String param1Value = filterConfig.getInitParameter(REPLACE_PARAM_NAME);
```
fhuaulme: Variable naming.

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/Utils.java
```diff
@@ -0,0 +1,50 @@
...
+
+import javax.servlet.http.HttpServletRequest;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.List;
+
+/**
+ *
+ */
+class Utils {
```
fhuaulme: Could you use a less generic/over-used name for this class?

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/Utils.java
```diff
@@ -0,0 +1,50 @@
...
+            putInMDCIfNotNull(HttpHeadersMDC.HTTP_REMOTE_ADDR, request.getRemoteAddr());
+        }
+        putInMDCIfNotNull(HttpHeadersMDC.HTTP_REMOTE_PORT, request.getRemotePort());
+        putInMDCIfNotNull(HttpHeadersMDC.HTTP_REMOTE_USER, request.getRemoteUser());
+        putInMDCIfNotNull(HttpHeadersMDC.HTTP_USER_AGENT, toString(request.getHeaders(USER_AGENT)));
+        putInMDCIfNotNull(HttpHeadersMDC.HTTP_FORWARDED_FOR, toString(request.getHeaders(X_FORWARDED_FOR)));
+    }
+
+    static void putInMDCIfNotNull(String key, Object value) {
+        if (value != null) {
```
fhuaulme: Or replace the `... != null` with a `String.valueOf(value)`

daikon-logging/logging-http-headers/src/test/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilterTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.net.InetSocketAddress;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ *
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = { HttpHeadersMDCFilterTest.TestApp.class }, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+public class HttpHeadersMDCFilterTest extends AbstractHttpHeadersMDCTest {
```
fhuaulme: It _seems_ to me a little overkill to start a full Spring context for testing a java.servlet.Filter. Did you consider using Mockito there?

daikon-logging/logging-http-headers/src/test/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCValveTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ *
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = { HttpHeadersMDCValveTest.TestApp.class }, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+public class HttpHeadersMDCValveTest extends AbstractHttpHeadersMDCTest {
```
fhuaulme: Same comment as before about the use of a full Spring context.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/96/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/96/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/97/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/97/)

daikon-logging/logging-http-headers/README.adoc
IevgenTarasov: Thanks for the review. Servlet filter is of course preferred over valve, and if you are deciding between the two then your choice should be definitely filter. Valve is there in case you cannot or do not want to use filter for some reason, but in that case you know what you're doing. I've added a line in the docu that filter is preferred. It's better to have something and not need it than the other way round.

daikon-logging/logging-http-headers/README.adoc
IevgenTarasov: Added it (I didnt have a dependency on spring before, that's why I didnt add it earlier)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/98/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/98/)

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDC.java
IevgenTarasov: I didnt want to add dependency on layout since the purpose of this lib has nothing to do with it (also this would add unwanted dependencies which I tried to avoid). Moreover, it's unlikely that anybody will access these values in MDC in runtime (but if somebody needs to do so this lib has only one mandatory dependency - slf4j-api, so its footprint is not too big).

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilter.java
IevgenTarasov: I agree it's a bit too long, but I couldnt come up with a shorter name which could still be self-explanatory. Generally it's not such an exceptional thing, spring boot has quite of few long parameter names https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html
spring.mvc.throw-exception-if-no-handler-found or spring.mobile.devicedelegatingviewresolver.enable-fallback

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/Utils.java
IevgenTarasov: I've renamed it

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/Utils.java
IevgenTarasov: This would produce a different result: now if value is null we dont put anything into the context. With `String.valueOf(value)` it would put the string "null" into MDC which was not desirable.

daikon-logging/logging-http-headers/src/test/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilterTest.java
IevgenTarasov: Proper context allows better testing for asynchronous stuff, and is probably more reliable (especially for the async calls). Also the test itself doesn't take too much time, even on the jenkins. Here you can see it took only about 7 seconds: https://ci-common.datapwn.com/job/daikon_pull-request_master/98/consoleFull

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilter.java
sgandon: I think Francois did not mean that the variable name was too long but more that it should be meaningfull.
if it represents REPLACE_PARAM_NAME, then why don't you call it replaceParamName ?

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/99/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/99/)

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilter.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        if (request instanceof HttpServletRequest) {
+            HttpServletRequest httpRequest = (HttpServletRequest) request;
+
+            MDCUtils.fillMDC(httpRequest, replaceRemoteAddrWithForwardedFor);
+
+            if (ClassUtils.isSpringAvailable()) {
```
rmannibucau: no need to synchronize the call on each request with a loadclass, this test can be moved to init()

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilter.java
```diff
@@ -0,0 +1,69 @@
...
+                SpringUtils.setAsyncMDCInterceptor(httpRequest, replaceRemoteAddrWithForwardedFor);
```
rmannibucau: can you do it for not spring case to wrapping the httpservletrequest to be able to wrap the async context. Also an executorservice wrapper is worth it by itself

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCValve.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public void setReplaceRemoteAddrWithForwardedFor(boolean replaceRemoteAddrWithForwardedFor) {
+        this.replaceRemoteAddrWithForwardedFor = replaceRemoteAddrWithForwardedFor;
+        LOG.debug("Setting replaceRemoteAddrWithForwardedFor to {}", replaceRemoteAddrWithForwardedFor);
+    }
+
+    @Override
+    public void invoke(Request request, Response response) throws IOException, ServletException {
+        if (request != null) {
```
rmannibucau: same as for the filter (springutils usage) + it doesn't support async case I think

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilter.java
IevgenTarasov: Thanks for the review. That's a good point, I will move the check to the constructor. Also in the comment above you've mentioned that "SpringUtils will create contentions if done in doFilter" - could you please elaborate on that? How exactly?

daikon-logging/logging-http-headers/src/main/java/org/talend/daikon/logging/http/headers/HttpHeadersMDCFilter.java
IevgenTarasov: Ah, oops... This makes sense, will change the var name to replaceParamName.

fhuaulme:
Looks ok to me for Spring integration. Maybe @rmannibucau comments could be addressed in another change set? (I'm suspecting you're afraid of integration in meecrowave?)

rmannibucau:
@fhuaulme not meecrowave - I doubt to use this one like that - but more any servlet container, including spring-boot servlet integration which would be broken. Ok to not support it for now for me

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/daikon_pull-request_master/105/) [:x:](https://ci-common.datapwn.com/job/daikon_pull-request_master/105/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/106/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/106/)

PR #340, opened 2018-02-13
fix(TMDM-11854): Output with keywords fields names

olhava:
* added utility methods for validation of field names

 Utility methods are necessary to validate the field names on Java conventions.
 Add utility methods to daikon, so that this functionality will be available in all components
**Link to the JIRA issue**

 https://jira.talendforge.org/browse/TMDM-11854

 
### DIFF
--- a/daikon/src/main/java/org/talend/daikon/avro/JavaNamesValidationHelper.java
+++ b/daikon/src/main/java/org/talend/daikon/avro/JavaNamesValidationHelper.java
@@ -1,0 +1,46 @@
+package org.talend.daikon.avro;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+/**
+ * Helper class that provides utility methods for validation of identifiers according to Java naming conventions.
+ */
+public class JavaNamesValidationHelper {
+
+    private static final Set<String> JAVA_KEYWORDS = new HashSet<String>(Arrays.asList("abstract", "continue", "for", "new",
+            "switch", "assert", "default", "goto", "package", "synchronized", "boolean", "do", "if", "private", "this", "break",
+            "double", "implements", "protected", "throw", "byte", "else", "import", "public", "throws", "case", "enum",
+            "instanceof", "return", "transient", "catch", "extends", "int", "short", "try", "char", "final", "interface",
+            "static", "void", "class", "finally", "long", "strictfp", "volatile", "const", "float", "native", "super", "while"));
+
+    private static final String CONTEXT_AND_VARIABLE_PATTERN = "^[a-zA-Z_][a-zA-Z_0-9]*$";
+
+    /**
+     * Checks whether specified name is a valid identifier according to Java conventions.
+     *
+     * @param name identifier to validate
+     * @return true, if name is a valid Java identifier
+     */
+    public static boolean isValidParameterName(String name) {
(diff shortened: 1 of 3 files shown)


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/117/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/117/)

daikon/src/main/java/org/talend/daikon/avro/JavaNamesValidationHelper.java
```diff
@@ -0,0 +1,46 @@
...
+     * @return true, if name is a valid Java identifier
+     */
+    public static boolean isValidParameterName(String name) {
+        if (name != null) {
+            if (isJavaKeyWord(name)) {
+                return false;
+            }
+            return Pattern.matches(CONTEXT_AND_VARIABLE_PATTERN, name);
+        }
+        return false;
```
dmytro-sylaiev: I believe you can simplify this method

daikon/src/main/java/org/talend/daikon/avro/JavaNamesValidationHelper.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public class JavaNamesValidationHelper {
+
+    private static final Set<String> JAVA_KEYWORDS = new HashSet<String>(Arrays.asList("abstract", "continue", "for", "new",
+            "switch", "assert", "default", "goto", "package", "synchronized", "boolean", "do", "if", "private", "this", "break",
+            "double", "implements", "protected", "throw", "byte", "else", "import", "public", "throws", "case", "enum",
+            "instanceof", "return", "transient", "catch", "extends", "int", "short", "try", "char", "final", "interface",
+            "static", "void", "class", "finally", "long", "strictfp", "volatile", "const", "float", "native", "super", "while"));
+
+    private static final String CONTEXT_AND_VARIABLE_PATTERN = "^[a-zA-Z_][a-zA-Z_0-9]*$";
```
gonchar-ivan: I suggest here
`private static final Pattern CONTEXT_AND_VARIABLE_PATTERN = Pattern.compile("^[a-zA-Z_][a-zA-Z_0-9]*$");`
Pattern compilation is considered as heavy operation, so it is worth to pre-compute it.

daikon/src/main/java/org/talend/daikon/avro/JavaNamesValidationHelper.java
gonchar-ivan: Dmytro is right, but it is better to provide some suggestions about simplifying. 
Nested code blocks (nested if-else and loop clauses) are considered as complicated. So, it is better to avoid them.
Also you may combine several conditions into one, when the same action is called in the `if` block.
E.g.
Conditions are
`name == null`
`isJavaKeyWord(name)'
action is
`return false`
it can be combined in following way:
```
if (name == null || isJavaKeyWord(name)) {
    return null;
}
```
Note, is name is null ` isJavaKeyWord(name)` won't be invoked.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/118/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/118/)

gonchar-ivan:
please, add some unit-tests

PR #354, opened 2018-03-21
feat(TDS-3154): support single quotes in TQL queries

afourmi:
The TQL support filters like : field = 'literal value'. The apostrophe ' is used as a delimiter, to let ANTLR find literal tokens. But if the literal itself contains an apostrophe, the TQL is not able do parse correctly the literal: depending of the whole filter parsed by TQL, either it leads to an exception being thrown during the parse operation, either the literal value is truncated and then the TQL results are not consistent with the filter.
 
- The TQL grammar now accepts literal with single quotes inside, eg _O'Reilly_ is now accepted. (see TQLParser.g4)
- After the parse operation, the grammar rules which use a literal value unescape single quotes from the literal inputs. (see TqlExpressionVisitor.java)
Only the apostrophe character is escaped, because it is the only character which is used by the TQL grammar, in the case of literal. 
 
**Link to the JIRA issue**
https://jira.talendforge.org/browse/TDS-3154
 
### DIFF
--- a/daikon-tql/daikon-tql-core/src/main/antlr4/TqlLexer.g4
+++ b/daikon-tql/daikon-tql-core/src/main/antlr4/TqlLexer.g4
@@ -53,4 +53,5 @@ DECIMAL: ('-')?('0' .. '9')+('.')('0' .. '9')*;
 FIELD: ('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | '_' | '.' )+;
 
 // literal values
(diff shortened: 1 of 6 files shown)


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/177/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/177/)

PR #515, opened 2019-06-17
fix(TUP-23482) remove direct dependency to slf4j

nrousseau:

This crypto utils is also used in the studio where many versions of slf4j might be needed by different component / cause conflict issues.
Best would be to not have any dependency to sl4fj (same for httpclient / StringUtils, but classes are not used in studio) 

Temporary fix to avoid SLF4J in the main class used by studio.
Need to be reviewed for a better solution, this is only a proposition to review / fix studio issue. 
**Link to the JIRA issue**
https://jira.talendforge.org/browse/TUP-23482
 
### DIFF
--- a/daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/Encryption.java
+++ b/daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/Encryption.java
@@ -8,8 +8,6 @@ import javax.crypto.Cipher;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.client.utils.URIBuilder;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * This class provides a helper class to encrypt and decrypt a given string using provided {@link CipherSource} and
@@ -17,7 +15,6 @@ import org.slf4j.LoggerFactory;
  */
 public class Encryption {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(Encryption.class);
 
     private final KeySource source;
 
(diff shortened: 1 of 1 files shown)


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/4/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/4/)

coheigea:
@nrousseau but daikon crypto doesn't have a dependency on slf4j-log4j as per the JIRA?

mvn dependency:tree | grep slf4j
[INFO] +- org.slf4j:slf4j-api:jar:1.7.25:compile

Just depending o the API should be fine IMO. If necessary you should just be able to exclude the API in the Studio if you have conflicting versions

nrousseau:
problem is that the crypto is added at the routine level, which is the base of every job built.
It becomes really difficult then to review the dependencies.
Even more some components need older / newer version of slf4j.
So the less dependencies we have, the best it is

coheigea:
What about if we change the scope of slf4j-api to provided? Would that work for you?

bhe-talendbj:
> What about if we change the scope of slf4j-api to provided? Would that work for you?

This does not work for us. Studio have thousands of components, each component have its dependencies, some component depends on slf4j, some component does not depend on it.

Studio generates jobs(code/pom/classpath etc) according to the components used, so if daikon changed scope to provided, then some jobs generated may does not have slf4j in classpath, this will result in failure of running job.

It seems removing slf4j from daikon is the best option for now.

Could you please help review above?

coheigea:
This PR can be closed I think...

PR #518, opened 2019-07-03
Re-use File object

coheigea:
Just a trivial merge to re-use a File object that we created with JarFile instead of passing in the raw file name.


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/daikon_pull-request_master/8/) [:star:](https://ci-common.datapwn.com/job/daikon_pull-request_master/8/)

PR #520, opened 2019-07-09
TDKN-252 - Support specifying Provider objects in CipherSources

coheigea:
This task is to support specifying Provider objects in CipherSources.

### DIFF
--- a/daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/CipherSources.java
+++ b/daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/CipherSources.java
@@ -4,6 +4,7 @@ import static org.talend.daikon.crypto.EncodingUtils.BASE64_DECODER;
 import static org.talend.daikon.crypto.EncodingUtils.BASE64_ENCODER;
 
 import java.security.Key;
+import java.security.Provider;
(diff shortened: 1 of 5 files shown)


daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/CipherSources.java
```diff
@@ -35,10 +36,17 @@ public static CipherSource getDefault() {
      * @return A {@link CipherSource} using AES encryption.
      */
     public static CipherSource aes() {
+        return aes(null);
+    }
+
+    /**
+     * @return A {@link CipherSource} using AES encryption and the given Provider
+     */
+    public static CipherSource aes(Provider p) {
```
fhuaulme: Can you please add a quick note in README of `crypto-utils` to highlight use cases where using Provider would be useful?

coheigea: Yep done! I don't think it will be very useful in general - it's just that I spotted some old encryption code in TAC I think, that uses the BouncyCastle Provider, so it makes it easier to migrate to use Daikon if we already have the support in place.

PR #523, opened 2019-07-10
feat(TDKN-255): configurable and lazy http backend

rmannibucau:

The http backend uses a - created eagerly - single default thread pool so:

1. it is hard to use in admin tools to filter the thread name,
2. it does not scale much (we commonly use ~8 threads for this kind of thing) - note it can need some bulk strategy later
3. if the pool is not used (error at bootstrap in the job) the the thread is still created for nothing
 
Make the init lazy, make it configurable through system properties and setters, and name the threads.
 
**Link to the JIRA issue**

 

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
```diff
@@ -1,22 +1,30 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
 import javax.xml.bind.DatatypeConverter;
 
 /**
- *
+ * todo: add bulk!
```
fhuaulme: Please create a Jira instead of adding a TODO comment.

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
```diff
@@ -78,7 +104,32 @@ public void setEncoding(String encoding) {
...
 
     public void sendEventAsync(String jsonEvent) {
-        executor.execute(new LogSender(jsonEvent));
+        ExecutorService executorService = executor.get();
+        if (executorService == null) {
+            final int maxSize = Integer.getInteger(getClass().getName() + ".max", this.maxSize);
+            executorService = new ThreadPoolExecutor(
+                    Math.min(Integer.getInteger(getClass().getName() + ".core", coreSize), maxSize), maxSize,
+                    Integer.getInteger(getClass().getName() + ".keepAliveMs", keepAliveMs), MILLISECONDS,
+                    new LinkedBlockingQueue<>(), // should it be bounded/synchronous or "boundable"?
```
fhuaulme: Create a queue size property (with a default value).

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
```diff
@@ -78,7 +104,32 @@ public void setEncoding(String encoding) {
     }
 
     public void sendEventAsync(String jsonEvent) {
-        executor.execute(new LogSender(jsonEvent));
+        ExecutorService executorService = executor.get();
+        if (executorService == null) {
+            final int maxSize = Integer.getInteger(getClass().getName() + ".max", this.maxSize);
```
fhuaulme: Update README with newly added system properties to configure this.

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
```diff
@@ -78,7 +104,32 @@ public void setEncoding(String encoding) {
...
+                        @Override
+                        public Thread newThread(final Runnable worker) {
+                            final Thread thread = new Thread(worker,
+                                    HttpEventSender.class.getName() + "-thread-" + counter.incrementAndGet());
+                            thread.setPriority(Thread.NORM_PRIORITY);
+                            thread.setDaemon(false);
+                            return thread;
+                        }
+                    });
+            if (!executor.compareAndSet(null, executorService)) {
```
fhuaulme: I'm a bit confused by this part: there might be concurrent calls to sendEventAsync(...), that may in turn lead to having 2 instances of executorService... but the first one will be shut down. What would happen if a LogSender was being executed by the first executor service?

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
rmannibucau: https://jira.talendforge.org/browse/TDKN-254

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
rmannibucau: The idea is 

1. if the executor is not set then create an instance
2. try to set the instance and if it succeeds then use it, if it fails then shutdown the just created instance (dead instance, never used) and get the one which succeeded in the other thread

so at the end you never have a case where you close an instance which can be used, no?

daikon-audit/README.adoc
```diff
@@ -96,6 +96,29 @@ input {
...
 
+==== HTTP Configuration
+
+HTTP appenders (logback and log4j1) share the following configuration:
+
+// org.talend.logging.audit.impl.http.HttpEventSender:documentation:start
+
+[opts="header",cols="a,a,a"]
+|====
+| Property | Description | Default Value
```
fhuaulme: Ok, looks slightly overkill... anyway, the actual system property name is `HttpEventSender.class.getName() + "." + field.getName()` and not only `field.getName()`

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
```diff
@@ -130,18 +211,57 @@ protected HttpURLConnection openConnection() {
...
-        executor.shutdownNow();
+    public void stop() {
+        final ExecutorService executorService = executor.get();
+        if (executor.compareAndSet(executorService, null)) {
+            executorService.shutdown();
+
+            try {
+                executorService.awaitTermination(30L, TimeUnit.SECONDS);
+            } catch (final InterruptedException e) {
+                Thread.currentThread().interrupt();
```
fhuaulme: Please add a log of the exception (in debug level)

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
fhuaulme: My understanding is the following may happen:

(t0) T1 creates an executor e1
(t1) T2 creates an executor e2
(t1) T1 calls execute(new LogSender(...)) in e1
(t2) T2 shuts down e1, and cause LogSender(...) to be discarded.

daikon-audit/README.adoc
rmannibucau: this is written and the doc goes with supportsSystemPropertiesOverride AFAIK

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
rmannibucau: There is no change in that logic - it is what is on master.
That said, i'm happy to add a logger, just pick one you are happy with, System.err?

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/http/HttpEventSender.java
rmannibucau: Think t2 (last line) is wrong, it shuts down e2 which is not used, gets back e1 then do the execute.

PR #533, opened 2019-08-07
feat(TDKN-256): Add file source for keys

fhuaulme:
 
There's a fixed key in CryptoHelper code.

* Implement a KeySource that gets key from file.
* Use file KeySource instead of hard coded key.

**Link to the JIRA issue**
https://jira.talendforge.org/browse/TDKN-256
 

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/KeySources.java
```diff
@@ -140,4 +149,54 @@ public static KeySource pbkDf2(String password, byte[] salt, int keyLength) {
...
+
+        private final String propertyName;
+
+        private FileSource(String propertyName) {
+            this.propertyName = propertyName;
+        }
+
+        @Override
+        public byte[] getKey() throws Exception {
+            final String o = String.valueOf(properties.get(propertyName));
```
coheigea: What if propertyName is not contained in properties here?

daikon-crypto/crypto-utils/src/test/java/org/talend/daikon/crypto/KeySourcesTest.java
```diff
@@ -47,6 +47,15 @@ public void shouldGenerateFromSystemProperty() throws Exception {
...
 
+    @Test
+    public void testGetKey() throws Exception {
+        final String systemKey = new String(KeySources.file("system.encryption.key").getKey(), EncodingUtils.ENCODING);
+        assertEquals("99ZwBDt1L9yMX2ApJx fnv94o99OeHbCGuIHTy22 V9O6cZ2i374fVjdV76VX9g49DG1r3n90hT5c1", systemKey);
+
+        final String propertiesKey = new String(KeySources.file("properties.encryption.key").getKey(), EncodingUtils.ENCODING);
+        assertEquals("Il faudrait trouver une passphrase plus originale que celle-ci!", propertiesKey);
+    }
+
```
coheigea: Maybe add a test here as well to check that we can add a new property from a file instead of reading from key.dat

daikon/src/main/java/org/talend/daikon/security/CryptoHelper.java
```diff
@@ -44,7 +46,16 @@
     // Iteration count
     private int iterationCount = 29;
 
-    public static final String PASSPHRASE = "99ZwBDt1L9yMX2ApJx fnv94o99OeHbCGuIHTy22 V9O6cZ2i374fVjdV76VX9g49DG1r3n90hT5c1"; //$NON-NLS-1$
+    public static final String defaultPassphrase;
+
```
coheigea: Changing from PASSPHRASE -> defaultPassphrase will break downstream code, maybe leave it as PASSPHRASE for now. Also, could you add a test to make sure that CryptoHelper.PASSPHRASE is equal to the old hard-coded String?

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/KeySources.java
fhuaulme: I'll put an exception (IllegalArgumentException)

daikon/src/main/java/org/talend/daikon/security/CryptoHelper.java
fhuaulme: 👍

PR #546, opened 2019-08-12
feat(encryption) Provide helpers to ease encryption migration

fhuaulme:

Applications sometimes need to change encryption of previously encrypted values (with different encryption configuration).

This PR adds 2 helper classes:

* EncryptionMigration: ease migration of values.
* PropertiesMigration: ease migration of properties file.
 
**Link to the JIRA issue**
https://jira.talendforge.org/browse/XXX (no jira yet, but maybe worth one).
 

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/PropertiesEncryption.java
```diff
@@ -43,7 +43,18 @@ public PropertiesEncryption(Encryption encryption) {
...
-        modifyAndSave(input, mustBeEncrypted, this::encryptIfNot);
+        encryptAndSave(input, mustBeEncrypted, this::encryptIfNot);
+    }
+
+    /**
+     * Reads the specified property file and then encrypts, if not already encrypted, the specified set of properties
+     * and overwrites the specified property file.
+     *
+     * @param input the property file
+     * @param mustBeEncrypted the set of properties that must be encrypted
```
coheigea: Missing javadoc for the third parameter

daikon-crypto/crypto-utils/src/test/java/org/talend/daikon/crypto/migration/EncryptionMigrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+        EncryptionMigration.build(new Encryption(KeySources.random(6), CipherSources.getDefault()), null);
+    }
+
+    @Test
+    public void shouldReencrypt() throws Exception {
+        // given
+        final Encryption source = new Encryption(KeySources.fixedKey("DataPrepIsSoCool"), CipherSources.aes());
+        final Encryption target = new Encryption(KeySources.random(16), CipherSources.getDefault());
+        final EncryptionMigration migration = EncryptionMigration.build(source, target);
+        final String originalEncrypted = "JP6lC6hVeu3wRZA1Tzigyg==";
```
coheigea: Why not just start with the plaintext, and encrypt it using source instead of hard-coding the encrypted value?

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/migration/EncryptionMigration.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Returns <code>true</code> is <code>encrypted</code> was created using {@link #migrate(String)} method,
+     * <code>false</code> otherwise.
+     * 
+     * @param encrypted A encrypted value either generated with a {@link Encryption}.
+     * @return <code>true</code> if encrypted value was generated by a call to {@link #migrate(String)}
+     * @see #migrate(String)
+     */
+    public boolean isMigrated(String encrypted) {
+        try {
```
coheigea: I'm a bit dubious about this method. If we're migrating a lot of values, the collection will build up pretty quickly. Also, "isMigrated" will return false if used with a new instance of the class, which may not be what the user expects. Is it really functionality that will be useful to us?

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/migration/EncryptionMigration.java
```diff
@@ -0,0 +1,114 @@
...
+     * 
+     * @param encrypted A encrypted value either generated with <code>source</code> or <code>target</code>
+     * {@link Encryption}.
+     * @return A decrypted value.
+     * @throws Exception Thrown in case of unexpected exception
+     * @see Encryption#decrypt(String)
+     * @see #isMigrated(String)
+     */
+    public String decrypt(String encrypted) throws Exception {
+        if (isMigrated(encrypted)) {
```
coheigea: I don't know if this method is really useful. The functionality of the class is to migrate, not to decrypt. If the user wants to decrypt they can use the target at a subsequent stage?

daikon-crypto/crypto-utils/src/test/java/org/talend/daikon/crypto/migration/EncryptionMigrationTest.java
fhuaulme: 👍

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/migration/EncryptionMigration.java
fhuaulme: Maybe not that useful after all. Removing it.

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/migration/EncryptionMigration.java
fhuaulme: Also removed as it was using removed method `isMigrated`

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/migration/EncryptionMigration.java
```diff
@@ -0,0 +1,62 @@
...
+        return new EncryptionMigration(source, target);
+    }
+
+    /**
+     * <p>
+     * This method uses <code>source</code> encryption to decrypt value and immediately encrypts value using
+     * <code>target</code>.
+     * </p>
+     * <p>
+     * It also remembers the values that were generated (keeps only a digest of the migration result, not the result
```
coheigea: Remove this comment about digests

PR #550, opened 2019-08-29
backport(TUP-23482) Remove direct dependency slf4j from daikon

bhe-talendbj:
Different version of slf4j are introduced in studio jobs.

Studio has thousands of components, each component has its dependencies, some component depends on slf4j, some component does not depend on it.

Studio generates jobs(code/pom/classpath/binary etc) according to the components used, so if daikon changed scope to provided, then some jobs generated may not have slf4j in classpath, this will result in failure of running job.

So it seems removing slf4j from daikon is the best option for now.

 
Remove direct dependency slf4j from daikon

 
**Link to the JIRA issue**
https://jira.talendforge.org/browse/TUP-23482
 
### DIFF
--- a/daikon-crypto/crypto-utils/pom.xml
+++ b/daikon-crypto/crypto-utils/pom.xml
@@ -12,11 +12,6 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
-            <version>${slf4j.version}</version>
-        </dependency>
-        <dependency>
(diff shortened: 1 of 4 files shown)


bhe-talendbj:
> Approved, but can you work with @nrousseau to change the PR for https://github.com/Talend/daikon/pull/550/files so that these changes get pushed to master as well?

Sure, I will work on https://github.com/Talend/daikon/pull/515 so that the same changes will be pushed into master.

Thank you for your review.

PR #580, opened 2019-12-23
feat(TDP-7886): standardize and upgrade Avro version

fabballe:
Avro version are not the same on all daikon project.
 
Standardize avro version and upgrade it to the last Avro version 

**Link to the JIRA issue**

 
### DIFF
--- a/daikon-messages/messages-model/src/main/resources/org/talend/daikon/messages/enum.vm
+++ b/daikon-messages/messages-model/src/main/resources/org/talend/daikon/messages/enum.vm
@@ -18,7 +18,6 @@
 #if ($schema.getNamespace())
 package $schema.getNamespace();
 #end
-@SuppressWarnings("all")
 #if ($schema.getDoc())
 /** $schema.getDoc() */
 #end
@@ -27,9 +26,9 @@ package $schema.getNamespace();
 #end
 @org.apache.avro.specific.AvroGenerated
 public enum ${this.mangle($schema.getName())} implements org.apache.avro.generic.GenericEnumSymbol<${this.mangle($schema.getName())}> {
-  #foreach ($symbol in ${schema.getEnumSymbols()})${this.mangle($symbol)}#if ($foreach.hasNext), #end#end
-  ;
-  public static final org.apache.avro.Schema SCHEMA$ = new org.apache.avro.Schema.Parser().parse("${this.javaEscape($schema.toString())}");
-  public static org.apache.avro.Schema getClassSchema() { return SCHEMA$; }
-  public org.apache.avro.Schema getSchema() { return SCHEMA$; }
+#foreach ($symbol in ${schema.getEnumSymbols()})${this.mangle($symbol)}#if ($foreach.hasNext), #end#end
+;
+public static final org.apache.avro.Schema SCHEMA$ = new org.apache.avro.Schema.Parser().parse("${this.javaEscape($schema.toString())}");
+public static org.apache.avro.Schema getClassSchema() { return SCHEMA$; }
+public org.apache.avro.Schema getSchema() { return SCHEMA$; }
 }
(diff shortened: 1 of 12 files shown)


daikon-messages/messages-model/src/main/resources/org/talend/daikon/messages/enum.vm
```diff
@@ -7,7 +7,7 @@
 ## "License"); you may not use this file except in compliance
```
RyanSkraba: Thanks for refreshing these templates as well!

You might want to copy these directly from https://github.com/apache/avro/blob/release-1.9.1/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/ without applying whitespace formatting -- this would make it easier to see the changes that were applied for the custom templates, and probably also gives better generated code.

(Even better, we should raise a PR in Avro to make the record base class configurable.  As far as I can tell, that's the only reason we don't use the templates built into Avro 1.9.1 directly.  What do you think?)

daikon-messages/messages-model/src/main/resources/org/talend/daikon/messages/record.vm
```diff
@@ -276,139 +378,466 @@ public class ${this.mangle($schema.getName())}#if ($schema.isError()) extends or
...
+throw new java.io.IOException("Corrupt ResolvingDecoder.");
+}
+}
+}
 }
+#end
+}
+
+#macro( encodeVar $indent $var $s )
+    #set ($I = $this.indent($indent))
```
RyanSkraba: Here's an example where the whitespace in the template messes with the whitespace for the generated code: `$I` should automatically be set to the leading whitespace necessary during code generation, and should be aligned to the left in the template: https://github.com/apache/avro/blob/release-1.9.1/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm#L597

daikon-messages/messages-model/src/main/resources/org/talend/daikon/messages/enum.vm
fabballe: Update from the source template DONE.

Indeed it can be a nice feature to make the record base class configurable

daikon-messages/messages-model/src/main/resources/org/talend/daikon/messages/record.vm
fabballe: Update from the source template DONE.

RyanSkraba:
On my end, I created https://issues.apache.org/jira/browse/AVRO-2712 (for Avro 1.10.0...)

RyanSkraba:
Hello!  Thanks for making the change!  Are you sure you copied the template classes from release-1.9.1?  They don't look consistent to me: https://github.com/apache/avro/blob/release-1.9.1/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/

Just for info -- Avro 1.9.2 should be coming out soon (within the month)!  It's mostly bug fixes, so I have no objection to doing a 1.9.1 version between the two.

fabballe:
@RyanSkraba : yes I copy the template classes from the link you provide me. Which inconsistent do you see on this file ?

RyanSkraba:
Hello!  Sorry for the delay!  I was noting the whitespace differences between https://raw.githubusercontent.com/Talend/daikon/bec48dee0b7cec3bf9f575dbb02a26f81b16b2b4/daikon-messages/messages-model/src/main/resources/org/talend/daikon/messages/record.vm and https://raw.githubusercontent.com/apache/avro/release-1.9.1/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm

This is a *very* minor nitpick and should **not** block merging this PR!  The only reason I bring it up is that an exact copy (including whitespace) would make it easier to compare between the two.

LGTM!

PR #589, opened 2020-03-11
TDKN-277 - Update Spring Boot to 2.1.13

coheigea:
This task is to update Spring Boot to 2.1.13 to pick up high level CVE fixes associated with Spring + Tomcat.

https://jira.talendforge.org/browse/TDKN-277

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,7 +41,7 @@
         <de.flapdoodle.embed.mongo.version>2.2.0</de.flapdoodle.embed.mongo.version>
         <mongo-java-server.version>1.11.1</mongo-java-server.version>
 
-        <spring-boot.version>2.1.10.RELEASE</spring-boot.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -41,7 +41,7 @@
         <de.flapdoodle.embed.mongo.version>2.2.0</de.flapdoodle.embed.mongo.version>
         <mongo-java-server.version>1.11.1</mongo-java-server.version>
 
-        <spring-boot.version>2.1.10.RELEASE</spring-boot.version>
+        <spring-boot.version>2.1.13.RELEASE</spring-boot.version>
```
fabballe: On TSBI BOM the version used is 2.1.10.RELEASE so we can't upgrade this version without upgrading the version inside TSBI BOM.

Generally daikon should used TSBI BOM in order to maintain these versions

PR #610, opened 2020-05-15
feat(TDKN-288): Add new "custom audit backend" feature to daikon-audit

anicolas:

This PR allows daikon-audit users to dynamically instantiate a custom audit backend using configuration properties (it was previously only possible through the API).

This enables more flexibility, since it allows to switch backend dynamically (for example in DI jobs for Observability enterprise project).

This feature introduces a new type of backend to daikon-audit: "custom", and a new "backend class name" property, so that the audit backend can be dynamically specified using properties.

**Why introduce this feature in an 1.x branch?**

Because we need this feature for the Observability enterprise project, which has some impacts on DI jobs. Those jobs are currently using the `1.8.0` version of daikon-audit and we don't want to impose a migration (with breaking changes) to daikon 2.x. It should be done at a later time.

But since this feature is useful, it will be backported to daikon 2.x and 3.x branches:

* Report to 2.x branch: https://jira.talendforge.org/browse/TDKN-289
* Report to 3.x branch: https://jira.talendforge.org/browse/TDKN-290

**Link to the JIRA issue**

* Main JIRA: https://jira.talendforge.org/browse/TDKN-287
* 1.x implementation: https://jira.talendforge.org/browse/TDKN-288
 
### DIFF
--- a/daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/DefaultAuditLoggerBase.java
+++ b/daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/DefaultAuditLoggerBase.java
@@ -60,6 +60,19 @@ public class DefaultAuditLoggerBase extends AbstractAuditLoggerBase {
             this.logger = loadBackend(LOG4J2_BACKEND, config);
             break;
 
+        case CUSTOM:
+            final String backendClass = AuditConfiguration.BACKEND_CLASS_NAME.getString(config);
+
+            if (backendClass == null) {
+                throw new IllegalArgumentException("Selected backend is " + backend + " and no backend class name has been specified");
+            } else if (!Utils.isClassPresent(backendClass)) {
+                throw new IllegalArgumentException("Selected backend is " + backend + " and its associated backend class "
+                        + backendClass + " is not available on classpath");
+            }
+
+            this.logger = loadBackend(backendClass, config);
+            break;
+
         default:
             throw new IllegalArgumentException("Unsupported backend " + backend);
         }
--- a/daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/DefaultAuditLoggerBase.java
+++ b/daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/DefaultAuditLoggerBase.java
@@ -64,9 +64,9 @@ public class DefaultAuditLoggerBase extends AbstractAuditLoggerBase {
             final String backendClass = AuditConfiguration.BACKEND_CLASS_NAME.getString(config);
 
             if (backendClass == null) {
-                throw new IllegalArgumentException("Selected backend is " + backend + " and no backend class name has been specified");
+                throw new IllegalArgumentException("Selected backend is '" + backend + "' and no backend class name has been specified");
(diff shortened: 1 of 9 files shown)


anicolas:
@fhuaulme the ability to change audit backend doesn't seem to be documented. Should I add some documentation or is it something we'll do at a later time (never? 😄 ).

anicolas:
There seems to be test failures on `daikon-tql-mongo` 🤔 

It seems it can't connect to MongoDB... I'll try to investigate but I'll probably need some help on this one.

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/DefaultAuditLoggerBase.java
```diff
@@ -60,6 +60,19 @@ public DefaultAuditLoggerBase(AuditConfigurationMap externalConfig) {
             this.logger = loadBackend(LOG4J2_BACKEND, config);
             break;
 
+        case CUSTOM:
+            final String backendClass = AuditConfiguration.BACKEND_CLASS_NAME.getString(config);
+
+            if (backendClass == null) {
+                throw new IllegalArgumentException("Selected backend is " + backend + " and no backend class name has been specified");
```
fhuaulme, line 67: Please change message to `"Selected backend is '" + backend + "' and...` (add the surrounding single quote: this eases to detect empty/whitespace issues in error messages)

daikon-audit/audit-common/src/test/java/org/talend/logging/audit/custom/CustomAuditBackend.java
```diff
@@ -0,0 +1,45 @@
...
+import org.slf4j.MDC;
+import org.talend.logging.audit.LogLevel;
+import org.talend.logging.audit.impl.AbstractBackend;
+import org.talend.logging.audit.impl.AuditConfigurationMap;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class CustomAuditBackend extends AbstractBackend {
```
fhuaulme, line 12: A quick javadoc would be nice here (with mention of the load of the custom class by its name).

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/DefaultAuditLoggerBase.java
anicolas: It makes sense, but I don't know if you noticed: I simply kept the existing wording used for other error cases in this method.

Would you like me to make the change anyway?

fhuaulme: I honestly didn't check the previous message (only saw your change). Yes, please do the change as it improves a bit the existing code.

daikon-audit/audit-common/src/test/java/org/talend/logging/audit/custom/CustomAuditBackend.java
anicolas: Good idea 👍

PR #644, opened 2020-08-24
TDKN-295 - DigestSources.sha256 not mixing salt with password

coheigea:
DigestSources.sha256 is not actually digesting the salt with the password...it is just digesting them separately and appending them, which defeats the purpose of using a salt. Thankfully, we are not actually using this anywhere.


fhuaulme (approved):
LGTM

PR #645, opened 2020-09-01
feat(code) :setup Github code scanning

sgandon:
testing the new GH code analysis
 
try it.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,62 @@
+name: "CodeQL"
+
+on:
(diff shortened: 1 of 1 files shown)


coheigea (approved):

fhuaulme:
Initial integration test is inconclusive. Closing PR to avoid consuming too many of GH actions minutes.

coheigea:
Note that GitHub code scanning is free for public repos, so I don't think it should have been consuming our minutes.
Looking at the run, it needs to be configured with Nexus credentials somehow, I guess via a secret.

PR #648, opened 2020-09-08
feat(TDKN-297): Add ability to disable message output

fhuaulme:

Audit event message processing (including variable substitution) can be a costly operation, especially in use cases when a very high throughput is expected and message is discarded.
 
* Add MDC field to configure whether message go to output or not.
* Update README documentation with additional information.

**Link to the JIRA issue**
https://jira.talendforge.org/browse/TDKN-297
 
- [x] Related design / discussions / pages (not in JIRA), if any, are all linked or available in the Pull Request

 
### DIFF
--- a/daikon-audit/README.adoc
+++ b/daikon-audit/README.adoc
@@ -250,10 +250,6 @@ NOTE: If application is running it needs to be restarted for this change to take
 
 === Disable audit message in audit event
 
-You may use constant `org.talend.logging.audit.impl.AbstractAuditLoggerBase.MDC_OUTPUT_MESSAGE` in the MDC to disable
-message output in case all interesting information is located in MDC.
+It is possible for an `AbstractBackend` to disable log messages if it doesn't need it (e.g. performance consideration).
 
-```
-Context ctx = ContextBuilder.create(MDC_OUTPUT_MESSAGE, "false").build();
-
-```
+To do that, the implementation of the `AbstractBackend` have to override the method `enableMessageFormat` in order to return `false` (default value is `true` - backward compatibility considerations).
--- a/daikon-audit/README.adoc
+++ b/daikon-audit/README.adoc
@@ -247,3 +247,13 @@ log.appender=none
 ----
 
 NOTE: If application is running it needs to be restarted for this change to take effect.
+
+=== Disable audit message in audit event
+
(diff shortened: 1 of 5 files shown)


anicolas:
Thanks for this! This looks good 👍 

Could you just tell me why you chose to go with the MDC attribute to disable the message output, instead of adding a new property in the `AuditConfiguration`?

pcadeau (approved):

YohannTalend (approved):

fhuaulme:
> Thanks for this! This looks good 👍
> 
> Could you just tell me why you chose to go with the MDC attribute to disable the message output, instead of adding a new property in the `AuditConfiguration`?

There's no link from the class where message is formatted with the `AuditConfiguration`. Adding the ability to use `AuditConfiguration` would require more in-depth refactoring of the lib (and challenging some choices done by the original developers of this lib).
But I agree this forces users of lib to set a specific MDC flag, cost of adding a line to set a MDC flag seems **now** to have lower cost than a full refactoring.

pierCo:
IMO, using MDC is not intuitive for a developer, because

> MDC (Mapped Diagnostic Context): The basic idea of Mapped Diagnostic Context is to provide a way to enrich log messages with pieces of information that could be not available in the scope where the logging actually occurs, but that can be indeed useful to better track the execution of the program.

:point_up: Nothing relative to configuration in the definition of MDC

Furthermore, MDC is attached to a thread (normally) so, do we really want to change configuration per thread? (I don't have a response, it is a question)

Have you also thought about the solution below?

* `constructor`,  2 constructors: 
   * one without parameter (ensure backward compatibility)
   * another with a parameter `AuditConfiguration` parameter OR a boolean `formatMessage` (as you want)
* `protected method` such as `protected boolean isMessageFormated()` overrided or not by the child's class (according configuration for example)

I try to bring my stone.

anicolas:
Using the event context for this would also mean that some Observability-specific code has to be introduced in the job code.

It turns out that **for the Observability use case**, we don't need the log message. 

But the current audit-related code in the jobs is completely Observability-agnostic. Tomorrow, one might decide to plug a new backend to the jobs, and that backend might need the log message.

I'm just speculating of course, but it would seem a better separation of concerns to use the `AuditConfigurationMap`.

But I can understand the need of not refactoring the entire library... 😉

fhuaulme:
Don’t hesitate to propose an alternative implémentation using your comments in this PR or an other PR as you prefer @antoine @pierre

PR #654, opened 2020-09-24
fix(TDC-4420): Fix incorrect value type inference for path expressions

fhuaulme:
 
There's an issue in value type inference (based on method return type). When path contain multiple elements, type is taken from first path element instead of last one.

* Make sure to infer type based on last part of path.
 
**Link to the JIRA issue**
https://jira.talendforge.org/browse/TDC-4420
 

lcelton-talend (approved):
Check was painful with a replacement of the jar before running an unit test but it's LGTM

PR #668, opened 2020-10-27
TPSVC-15325: clean MDC context

vgagara-talend:
 https://jira.talendforge.org/browse/TPSVC-15325

clean up MDC properties after sending log 

**Link to the JIRA issue**

 
### DIFF
--- a/daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/SimpleAuditLoggerBase.java
+++ b/daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/SimpleAuditLoggerBase.java
@@ -62,8 +62,11 @@ public class SimpleAuditLoggerBase implements AuditLoggerBase {
     @Override
     public void log(LogLevel level, String category, Context context, Throwable throwable, String message) {
         Context actualContext = context == null ? ContextBuilder.emptyContext() : ContextBuilder.create(context).build();
-        backend.setNewContext(actualContext);
-
-        this.backend.log(category, level, message, throwable);
+        try {
+            backend.setNewContext(actualContext);
+            this.backend.log(category, level, message, throwable);
+        } finally {
+            backend.resetContext(null);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/SimpleAuditLoggerBase.java
```diff
@@ -62,8 +62,11 @@ private static AbstractBackend loadBackend(String className, AuditConfigurationM
     @Override
     public void log(LogLevel level, String category, Context context, Throwable throwable, String message) {
         Context actualContext = context == null ? ContextBuilder.emptyContext() : ContextBuilder.create(context).build();
-        backend.setNewContext(actualContext);
-
-        this.backend.log(category, level, message, throwable);
+        try {
```
fhuaulme, line 65: Wouldn't it be better to get backend's current context, store it in a "previousContext" variable, and do a `backend.resetContext(previousContext)` in the finally clause.

fhuaulme (changes requested):

daikon-audit/audit-common/src/main/java/org/talend/logging/audit/impl/SimpleAuditLoggerBase.java
vgagara-talend: originally I thought about the option to add current context to existing one (with MDC.put(...) in setContextMap() method) and later do MDC.remove() to only properties we added here. But this will change current behavior because currently we unconditionally "wipe" existing context here https://github.com/vgagara-talend/daikon/blob/master/daikon-audit/audit-kafka/src/main/java/org/talend/logging/audit/kafka/KafkaBackend.java#L84 and maybe it is intentional. Do you propose to add new method to `AbstractBackend` something like `getContextMap()` which will return current MDC context?

fhuaulme: It's better to stay consistent with KafkaBackend for the time being.

fhuaulme (approved):

fhuaulme (approved):

PR #674, opened 2020-11-26
Removing internal sun APIs from the CertificateGenerater

coheigea:
@RyanSkraba reported to me that the signature verifier tests fail on his machine. I think this is due to the fact that the tests use the internal Sun APIs. Instead I changed the tests to generate certificates using BouncyCastle instead, which means it should run on any JVM.


fhuaulme (approved):
LGTM

PR #696, opened 2020-12-24
TPSVC-16764: [LOGGING] Make daikon-logging ECS compatible

jehervy:
In the scope of [reliability track - logging topic](https://jira.talendforge.org/browse/PMGT-328),  the way to migrate Talend services to the common logging format [ECS (Elastic Common Schema)](https://www.elastic.co/guide/en/ecs/current/index.html), powered by Elastic, has been studied.

For that purpose, it has been decided to leverage the library `daikon-logging`.

The goal of this pull request is to make `daikon-logging` **100% ECS compatible**.

Please read [updated README](https://github.com/Talend/daikon/tree/jhervy/feat/tpsvc-16754-integrate-ecs/daikon-logging/logging-event-layout) for detailed information.

The following main changes have been done :

## Leverage [`ecs-logging-java`](https://github.com/elastic/ecs-logging-java) lib

Logback & Log4j2 have been rewritten by leveraging [`ecs-logging-java`](https://github.com/elastic/ecs-logging-java) core lib.

It helped to deal with the serialization of [the base ECS fields](https://www.elastic.co/guide/en/ecs/current/ecs-base.html):

- [`@timestamp`](https://www.elastic.co/guide/en/ecs/current/ecs-base.html#field-timestamp)
- [`labels`](https://www.elastic.co/guide/en/ecs/current/ecs-base.html#field-labels)
- [`message`](https://www.elastic.co/guide/en/ecs/current/ecs-base.html#field-message)
- [`tags`](https://www.elastic.co/guide/en/ecs/current/ecs-base.html#field-tags)

and some extended ones like:

- [`ecs`](https://www.elastic.co/guide/en/ecs/current/ecs-ecs.html)
- [`log`](https://www.elastic.co/guide/en/ecs/current/ecs-log.html)
- [`process`](https://www.elastic.co/guide/en/ecs/current/ecs-process.html)
- [`error`](https://www.elastic.co/guide/en/ecs/current/ecs-error.html).

A custom serializer has been implemented in order to handle the ECS extended fields 

- [`host`](https://www.elastic.co/guide/en/ecs/current/ecs-host.html)
- [`event`](https://www.elastic.co/guide/en/ecs/current/ecs-event.html)

## ECS version

The version of ECS used by the current library is referenced in the file [`logging.properties`](https://github.com/Talend/daikon/blob/jhervy/feat/tpsvc-16754-integrate-ecs/daikon-logging/logging-event-layout/src/main/resources/logging.properties)  : 

`logging.properties`
```properties
ecs.version=1.7.0
```

This version will be : 

- Loaded by maven during `INITIALIZE` phase in order to retrieve the [`ecs_flat.yml` file provided by Elastic](https://github.com/elastic/ecs/blob/master/generated/ecs/ecs_flat.yml).
- Injected in the generated logs.

## Generating ECS fields enum

In order to help the developers to look for an ECS field or even to provide a way for the current lib to filter out non ECS fields, the [`ecs_flat.yml` file provided by Elastic](https://github.com/elastic/ecs/blob/master/generated/ecs/ecs_flat.yml) is pulled during maven `GENERATE-SOURCES` phase (the `ecs.version` property is used to retrieve the corresponding file).

The `ecs_flat.yml` file is then converted, during the same maven phase, to a java enumeration using the freemarker template [`ecs_flat.ftl`](https://github.com/Talend/daikon/blob/jhervy/feat/tpsvc-16754-integrate-ecs/daikon-logging/logging-event-layout/src/main/resources/ecs_flat.ftl).

Extract from `ecs_flat.yml`
```yaml
'@timestamp':
  dashed_name: timestamp
  short: Date/time when the event originated.
  type: date
agent.build.original:
  dashed_name: agent-build-original
  short: Extended build information for the agent.
  type: wildcard
```

`ecs_flat.ftl`
```ftl
package org.talend.daikon.logging.ecs;

<#assign className = output.file.name?keep_before(".")>
<#assign fields = model.content>

/**
* Enumeration of all ECS fields based on ecs_flat.yaml file
*/
public enum ${className} {
<#list fields?keys as key>
    /**
    * ${fields[key].short}
    * Type: ${fields[key].type}
    <#if fields[key].example??>
        <#if fields[key].example?is_date>
    * Example: ${fields[key].example?datetime_if_unknown?iso_utc}
        <#elseif fields[key].example?is_number>
    * Example: ${fields[key].example?c}
        <#else>
    <#-- Remove all non ASCII characters -->
    * Example: ${fields[key].example?replace("[^\\p{ASCII}]", "", "r")}
        </#if>
    </#if>
    */
    ${fields[key].dashed_name?replace('-', '_')?upper_case}("${key}")<#if key?is_last>;<#else>,</#if>
</#list>

    public final String fieldName;

    private ${className}(String fieldName) {
        this.fieldName = fieldName;
    }
}
```

Extract from `EcsFields.java`
```java
package org.talend.daikon.logging.ecs;


/**
* Enumeration of all ECS fields based on ecs_flat.yaml file
*/
public enum EcsFields {
    /**
    * Date/time when the event originated.
    * Type: date
    * Example: 2016-05-23T08:05:34.853Z
    */
    _TIMESTAMP("@timestamp"),
    /**
    * Extended build information for the agent.
    * Type: keyword
    * Example: metricbeat version 7.6.0 (amd64), libbeat 7.6.0 [6a23e8f8f30f5001ba344e4e54d8d9cb82cb107c built 2020-02-05 23:10:10 +0000 UTC]
    */
    AGENT_BUILD_ORIGINAL("agent.build.original"),
    // [...]
}
```

## Mapping & filtering MDC
 
As main goal of this lib is to make the generated logs 100% ECS compatible, we must ensure that all the output fields are ECS fields. For that purpose, the process described below is followed :

MDC mapping & filtering process

### Step1: MDC to ECS mapping

The first step is to map the MDC keys with the corresponding ECS fields by leveraging the [`mdc_ecs_mapping.properties`](https://github.com/Talend/daikon/blob/jhervy/feat/tpsvc-16754-integrate-ecs/daikon-logging/logging-event-layout/src/main/resources/mdc_ecs_mapping.properties) file.

`mdc_ecs_mapping.properties`
```yaml
accountId=cloud.account.id
executionId=labels.execution_id
stackName=labels.stack_name
userId=user.id
user-activity-id=labels.user_activity_id
X-talend-user-activity-id=labels.x_talend_user_activity_id
category=service.type
eventType=event.action
audit=labels.audit
X-B3-SpanId=span.id
X-B3-TraceId=trace.id
X-Span-Export=labels.span_export
```

### Step 2: ECS fields filtering

Once the fields mapped, all the non ECS fields are filtered out by leveraging the `EcsFields.java` enumeration generated earlier.

**Link to the JIRA issue**
https://jira.talendforge.org/browse/TPSVC-16764
 

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Check if a given field is an ECS label
+     * 
+     * @param field Field to check
+     * @return true if the field is an ECS label or false otherwise
+     */
+    public static boolean isECSLabel(String field) {
+        return field != null && field.split("\\.").length == 2 && LABELS_FIELD.equals(field.split("\\.")[0]);
```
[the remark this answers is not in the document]
pierCo: I found the answer, its type is `keyword` so no it is not possible to have hierarchical entries.

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Check if a given field is an ECS field (ECS label included)
+     * 
+     * @param field Field to check
+     * @return true if the field is an ECS field or false otherwise
+     */
+    public static boolean isECSField(String field) {
+        return getECSFields().contains(field) || isECSLabel(field);
```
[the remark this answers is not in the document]
pierCo: This case can be extended to all fields typed `keyword`.
Certainly out of scope for this version.

daikon-logging/logging-event-layout/src/main/resources/ecs_flat.ftl
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+* Constants for all ECS fields based on ecs_flat.yaml file
+*/
+public final class ${className} {
+
+<#list fields?keys as key>
+    /**
+    * ${fields[key].short}
+    * Type: ${fields[key].type}
```
pierCo, line 14: Adding `example` fielf could be great.
Is it possible?

daikon-logging/logging-event-layout/src/main/resources/ecs_flat.ftl
```diff
@@ -0,0 +1,19 @@
+package org.talend.daikon.logging.ecs;
+
+<#assign className = output.file.name?keep_before(".")>
+<#assign fields = model.content>
+
+/**
+* Constants for all ECS fields based on ecs_flat.yaml file
+*/
+public final class ${className} {
```
pierCo, line 9: In my mind, if I had to do that I would have made an `Enum` instead a set of constants.
Because:
- use it in method headers to force input param to the valid ECS field (more restrictive than "String")
- we can add more info and logic in an `enum` such field type or value validation according the type, ...
But: how to manage `keyword` fields with that?

But certainly out of scope for this version.

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
```diff
@@ -0,0 +1,78 @@
...
+        return INSTANCE;
+    }
+
+    /**
+     * Return the set of all ECS fields
+     * 
+     * @return the set of all ECS fields
+     */
+    public static Set<String> getECSFields() {
+        return getInstance().fields;
```
pierCo, line 56: It could be great to return a defensive copy of this `Set` (or an unmodifiable `Set`), WDYT?

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/MdcEcsMapper.java
```diff
@@ -0,0 +1,65 @@
...
+        return INSTANCE;
+    }
+
+    /**
+     * Return the MDC to ECS map
+     * 
+     * @return the MDC to ECS map
+     */
+    public static Map<String, String> getMapping() {
+        return getInstance().mapping;
```
pierCo, line 52: Defensive copy or an unmodifable `Map`, WDYT?

pierCo (commented):
👍

daikon-logging/logging-event-layout/pom.xml
```diff
@@ -80,4 +90,51 @@
 			<scope>test</scope>
 		</dependency>
     </dependencies>
+	<build>
+		<plugins>
+			<plugin>
```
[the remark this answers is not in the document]
jehervy: I see your point, I thought about generating the file and storing it in GIT but finally this way to do seems to simplify a lot the maintenance and the update.

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
```diff
@@ -0,0 +1,78 @@
...
+ * - whether a given field is an ECS label or not, based on pattern "labels.*"
+ * - whether a given field is an ECS field or not, based on ecs_flat.yml file
+ */
+public class EcsFieldsChecker {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(EcsFieldsChecker.class);
+
+    private static final String ECS_FLAT_FILE = "ecs_flat.yml";
+
+    private static final String LABELS_FIELD = "labels";
```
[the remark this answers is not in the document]
jehervy: Yes I'll add it 👍

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
```diff
@@ -0,0 +1,78 @@
...
+        ObjectMapper om = new ObjectMapper(new YAMLFactory());
+        try {
+            fields = om.readValue(ecsFlatFile, Map.class).keySet();
+        } catch (IOException e) {
+            LOGGER.error("ECS fields file can't be read", e);
+        }
+    }
+
+    private static EcsFieldsChecker getInstance() {
+        if (INSTANCE == null) {
```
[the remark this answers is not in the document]
jehervy: You are right 👍

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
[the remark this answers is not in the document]
jehervy: yes indeed, no hierarchical entries.

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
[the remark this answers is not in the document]
jehervy: Yes indeed, I'll update the condition as you suggested 👌

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/MdcEcsMapper.java
```diff
@@ -0,0 +1,65 @@
...
+        File mdcEcsMappingFile = new File(classLoader.getResource(MDC_ECS_MAPPING_FILE).getFile());
+        ObjectMapper om = new ObjectMapper(new YAMLFactory());
+        try {
+            mapping = (Map<String, String>) om.readValue(mdcEcsMappingFile, Map.class);
+        } catch (IOException e) {
+            LOGGER.error("MDC ECS mapping file can't be read", e);
+        }
+    }
+
+    private static MdcEcsMapper getInstance() {
```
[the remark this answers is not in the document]
jehervy: 👍

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/MdcEcsMapper.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Map a MDC key with its corresponding ECS field
+     * In case where not mapping exists, the MDC key is returned
+     * 
+     * @param mdcKey MDC key to map
+     * @return the corresponding ECS field or the MDC key if no mapping exists
+     */
+    public static String map(String mdcKey) {
+        return Optional.ofNullable(getMapping().get(mdcKey)).orElse(mdcKey);
```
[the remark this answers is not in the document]
jehervy: Let's make this expression more readable then 👌 I'll use `if/else` instead

daikon-logging/logging-event-layout/src/main/resources/ecs_flat.ftl
jehervy: I guess yes !

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
jehervy: 👌

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/MdcEcsMapper.java
jehervy: 👌

daikon-logging/logging-event-layout/src/main/resources/ecs_flat.ftl
jehervy: I don't see any objection to generating an enum instead of constants. 
You're right about the fact that it can be used as input parameter.
I'll make the change 👌

pierCo: As you want, but we have to pay attention to the fields typed `keyword` such as `labels`. How to manage that with an `enum` :)

jehervy: Type `keyword` is the basic string type.
I guess you mean fields with `type: object` and `object_type: keyword`.
Only 2 fields are concerned: `labels` and `container.labels`.
They can't be managed as enum but, as they are particular fields, I guess we can live with that.

pierCo: You are right :)
> fields with type: object and object_type: keyword.
Sorry for this approximation.

So my question stay opened 😄 -> how to manage that? A dedicated method to manage labels?

jehervy: Currently the `EcsFieldsChecker` is compatible with `labels`, let's update it to handle `container.labels` as well (even if this field is supposed to be filled later in the process).
Do you see something else ?

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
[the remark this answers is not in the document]
pierCo: Erratum: fields typed `object keyword` and not only `keyword`

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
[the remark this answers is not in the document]
jehervy: Even better: leverage generated `EcsFields` enum

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/event/layout/Log4j2JSONLayout.java
```diff
@@ -1,233 +1,257 @@
...
-        JSONObject logstashEvent = new JSONObject();
-        JSONObject userFieldsEvent = new JSONObject();
-        HostData host = new HostData();
+    public String getContentType() {
+        return "application/json";
+    }
 
-        // Extract and add fields from log4j2 config, if defined
-        LayoutUtils.addUserFields(additionalAttributes, userFieldsEvent);
+    private StringBuilder toText(LogEvent event, StringBuilder builder, boolean gcFree) {
```
sgandon, line 78:
```suggestion
    private StringBuilder toText(LogEvent event, StringBuilder builder) {
```
gcFree seems not to be used. (of course the call to this method should also be changed should you access this suggestion.

jehervy: Indeed good catch 👍

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/event/layout/Log4j2JSONLayout.java
```diff
@@ -1,233 +1,257 @@
...
+        EcsJsonSerializer.serializeFormattedMessage(builder, event.getMessage().getFormattedMessage());
+        EcsJsonSerializer.serializeEcsVersion(builder);
+        EcsJsonSerializer.serializeServiceName(builder, serviceName);
+        EcsJsonSerializer.serializeThreadName(builder, event.getThreadName());
+        EcsJsonSerializer.serializeLoggerName(builder, event.getLoggerName());
 
-        Map<String, String> mdc = LayoutUtils.processMDCMetaFields(loggingEvent.getContextData().toMap(), logstashEvent,
-                metaFields);
+        // Serialize custom markers with format key:value
+        serializeCustomMarkers(builder, event.getMarker());
```
sgandon, line 88: I am not sure to understand the difference between Markers and CustomMarkers, can you give some explanation please ?

daikon-logging/logging-event-layout/src/main/resources/ecs_flat.yml
```diff
@@ -0,0 +1,9262 @@
+'@timestamp':
```
sgandon, line 1: how do we know which version of ECS is used here ?

jehervy: It comes from [`ecs-logging-java`](https://github.com/elastic/ecs-logging-java) lib.
I guess it will be better to manage it on our side. 
We can specify it in a property file, or maybe you have better idea ?

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
```diff
@@ -0,0 +1,85 @@
...
+    // ECS Labels fields have a particular behavior as they contain custom keyword fields
+    private static final List<String> LABELS_FIELDS = Arrays.asList( //
+            EcsFields.LABELS.fieldName, //
+            EcsFields.CONTAINER_LABELS.fieldName //
+    );
+
+    private Set<String> fields = new HashSet<>();
+
+    private EcsFieldsChecker() {
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
```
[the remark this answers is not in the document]
jehervy: Indeed you're right, I'll made the change.
Actually, initially it was not an enum but a constants file. 
Now it's an enum we can totally leverage it here 👍

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
```diff
@@ -0,0 +1,85 @@
...
+        File ecsFlatFile = new File(classLoader.getResource(ECS_FLAT_FILE).getFile());
+        ObjectMapper om = new ObjectMapper(new YAMLFactory());
+        try {
+            fields = om.readValue(ecsFlatFile, Map.class).keySet();
+        } catch (IOException e) {
+            LOGGER.error("ECS fields file can't be read", e);
+        }
```
sgandon, line 42:
```suggestion
EnumSet.allOf(EcsFields.class).stream().forEach(ecsField -> fields.add(ecsField.fieldName));
```
yep something like this may be.

daikon-logging/logging-event-layout/src/main/resources/ecs_flat.yml
sgandon: Since it is not loaded dynamically anymore, there is not need to store this file in the repo, from what I understand this is now only used for the source generation so I guess it could be extracted on the fly during the source generation step.
What do you think ?

jehervy: Sure 👍 Good idea. I will do that 👌

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsChecker.java
pierCo: In addition, this will allow us to no longer depend on the YAML parser 🎊

sgandon:
@jehervy thanks for these updates. I have a few questions.
1/ I am wondering why you have to serialize the version as I thought is was serialzed by the library ?
2/ if this is because we may want to use a different version of the ECS spec, isn't there a possible version clash between the one we want the support and the one supported in the ecs java library ?

jehervy:
> @jehervy thanks for these updates. I have a few questions.
> 1/ I am wondering why you have to serialize the version as I thought is was serialzed by the library ?
> 2/ if this is because we may want to use a different version of the ECS spec, isn't there a possible version clash between the one we want the support and the one supported in the ecs java library ?

1/ Unfortunately, the lib serilizes it in [a hard coded way](https://github.com/elastic/ecs-logging-java/blob/master/ecs-logging-core/src/main/java/co/elastic/logging/EcsJsonSerializer.java#L52). If we want to manage the version on our side, we also have to serialize it.
2/ The [`ecs-logging-code`](https://github.com/elastic/ecs-logging-java/blob/master/ecs-logging-core) lib is used only to serialize base ECS fields. These fields shoulnd't change a lot with ECS version (they didn't change between lib ECS version - `1.2.0` - and current version - `1.7.0`).

jehervy:
> @jehervy thanks for these updates. I have a few questions.
> 1/ I am wondering why you have to serialize the version as I thought is was serialzed by the library ?
> 2/ if this is because we may want to use a different version of the ECS spec, isn't there a possible version clash between the one we want the support and the one supported in the ecs java library ?

@sgandon The more it goes, the more I have the feeling that we can get ride of [`ecs-logging-java`](https://github.com/elastic/ecs-logging-java) lib 🤔 

I initially planned to override the provided layouts before realizing that was not possible.
Then I wanted to reuse only the core serializer before realizing that our own must be implemented to fit out needs.

Currently we only use `ecs-logging-java` core serializer for "basic" ecs fields. But I'm wondering if we shouldn't deal with them by ourself.

sgandon:
> > @jehervy thanks for these updates. I have a few questions.
> > 1/ I am wondering why you have to serialize the version as I thought is was serialzed by the library ?
> > 2/ if this is because we may want to use a different version of the ECS spec, isn't there a possible version clash between the one we want the support and the one supported in the ecs java library ?
> 
> @sgandon The more it goes, the more I have the feeling that we can get ride of [`ecs-logging-java`](https://github.com/elastic/ecs-logging-java) lib 🤔
> 
> I initially planned to override the provided layouts before realizing that was not possible.
> Then I wanted to reuse only the core serializer before realizing that our own must be implemented to fit out needs.
> 
> Currently we only use `ecs-logging-java` core serializer for "basic" ecs fields. But I'm wondering if we shouldn't deal with them by ourself.

After some live discussion it is more clear now. I though the lib was linked to an ECS version but the link is weak as the ecs core lib does not even embed the esc spec.
So I believe we can still keep it as it is very light and provide some helpers method for serializing special types.

sgandon (approved):
LGTM

PR #747, opened 2021-05-24
TDKN-310 - Don't use CryptoHelper to encrypt properties in StringProp…

coheigea:
In StringProperty, the encryptStoredValue method uses CryptoHelper for both encryption and decryption. This task is to remove the "encrypt" part of this functionality, as in components it already overrides this:

https://github.com/Talend/components/blob/39419b8192721805e09bade037fd5104ee162289/core/components-api/src/main/java/org/talend/components/api/properties/ComponentBasePropertiesImpl.java#L33


fhuaulme:
@bhe-talendbj Did you have time to take a look at this PR?

bhe-talendbj (approved):

fhuaulme (approved):

PR #751, opened 2021-06-03
TDKN-312 - Make iteration count for pbkdf2 configurable

coheigea:
Currently, the iteration count for pbkdf2 is hard-coded at 65536. However, the OWASP password storage cheatsheet recommends:

https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

If FIPS-140 compliance is required, use PBKDF2 with a work factor of 310,000 or more and set with an internal hash function of HMAC-SHA-256.
We should make the iteration count configurable and recommend in Javadoc that an iteration count of 310000 is recommended.

I also did some re-factoring, to pull the encoding out of the Digester class and into PBKDF2, to make it easier to add new PasswordDigests (coming soon)


daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/KeySources.java
```diff
@@ -124,24 +124,46 @@ public static KeySource systemProperty(String systemProperty) {
...
      */
-    public static KeySource pbkDf2(String password, byte[] salt, int keyLength) {
+    public static KeySource pbkDf2(String password, byte[] salt, int keyLength, int iterations) {
         return () -> {
             if (salt == null || salt.length == 0) {
                 throw new IllegalArgumentException("Cannot use pbkDf2 with empty or null salt.");
             }
             try {
-                KeySpec spec = new PBEKeySpec(password.toCharArray(), salt, 65536, keyLength);
+                KeySpec spec = new PBEKeySpec(password.toCharArray(), salt, iterations, keyLength);
```
Gby56, line 166: Could we throw an exception here if the iterations are really too low, to have a cutoff/threshold/hard requirement in place

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/KeySources.java
```diff
@@ -124,24 +124,46 @@ public static KeySource systemProperty(String systemProperty) {
...
+     * <p>
+     * As recommendation, you may initialize a salt using {@link org.talend.daikon.crypto.KeySources#random(int)}.
+     * </p>
+     * 
+     * @deprecated Use {@link #pbkDf2(String, byte[], int, int)} instead with a stronger iteration count.
+     * @return A {@link KeySource} implementation using PBKDF2 and provided <code>salt</code>
+     */
+    @Deprecated
+    public static KeySource pbkDf2(String password, byte[] salt, int keyLength) {
+        return pbkDf2(password, salt, keyLength, 65536);
```
Gby56, line 141: On top of `@Deprecated` we could add a warning log maybe ? Otherwise it's only seen at compile time

Gby56 (changes requested):

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/KeySources.java
coheigea: Yes good idea. Say 50000? No one is using it with less than this anyway at Talend

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/KeySources.java
coheigea: Only the TDQ team are using it, I think instead I'd prefer to just make them migrate to the newer method once we release it. I don't really want to see warning logs in the Studio for it.

Gby56: Sure then yeah, it's cleaner that way

daikon-crypto/crypto-utils/src/main/java/org/talend/daikon/crypto/KeySources.java
Gby56: 60K and we've got a deal ! 🤝

Gby56 (approved):
Looks good to me !

fhuaulme (approved):

PR #767, opened 2021-08-31
TDKN-317 - Update Maven version to 3.8.2 to fix a CVE

coheigea:
https://sca.analysiscenter.veracode.com/workspaces/dppFqb3/issues/vulnerabilities/82573030

maven uses insecure TLS configurations. Non-SSL repository references are allowed by default and would potentially allow for man-in-the-middle attacks.

9.1 high - CVE-2021-26291

The fix is to update to 3.8.2


daikon-documentation/pom.xml
```diff
@@ -21,7 +21,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <maven.version>3.6.3</maven.version> <!-- latest available maven version in builders -->
+        <maven.version>3.8.2</maven.version>
```
fhuaulme, line 24: The image used to spawn K8S builder pods (in `Jenkinsfile`) needs to be updated as well with a image that provides Maven 3.8.2

fhuaulme (changes requested):

daikon-documentation/pom.xml
coheigea: OK let's wait for TSBI. For future reference, where does this need to be updated?

fhuaulme: It needs to be updated in 2 different places:
* https://github.com/Talend/daikon/blob/3e5de6627e087e5a67ebdbf456b1cd6761b7ea14/Jenkinsfile#L40 for the build
* https://github.com/Talend/daikon/blob/3e5de6627e087e5a67ebdbf456b1cd6761b7ea14/Jenkinsfile-srcclr#L17 for the veracode static checks

fhuaulme (approved):

PR #780, opened 2021-11-12
fix daikon for java 17, now only miss javascript engine

wwang-talend:
https://jira.talendforge.org/browse/TDI-46665
 
https://jira.talendforge.org/browse/TDI-46665

 

pom.xml
```diff
@@ -203,7 +203,7 @@
                 <groupId>com.cedarsoftware</groupId>
                 <artifactId>json-io</artifactId>
                 <!-- using a specific Talend artifact version until the next official release, the PR has been merge into master but not yet release. -->
-                <version>4.9.9-TALEND</version>
+                <version>4.13.0</version>
```
wwang-talend, line 206: will use next version, here only for local test, have merge that : https://github.com/jdereg/json-io/commit/3fea13ab41fe63aa2d4df1641b9d94c7979f8399

daikon/src/test/java/org/talend/daikon/number/BigDecimalFormatterTest.java
```diff
@@ -34,8 +35,9 @@ public void testToBigDecimalEU() throws Exception {
 
     @Test
     public void testToBigDecimalGroupingEU() throws Exception {
-        assertEquals("4" + ((char) 160) + "512,5",
-                BigDecimalFormatter.format(new BigDecimal("4512.50"), BigDecimalParser.EU_DECIMAL_PATTERN));
+        // see https://kirillbelyaev.com/s/
```
sgandon, line 38: I don't understand why you have changed this test ?
The space is the grouping separator and this has not changed.
Please have a

daikon/src/main/java/org/talend/daikon/runtime/RuntimeUtil.java
```diff
@@ -91,33 +92,23 @@ protected static void registerMavenUrlFactory() {
         // If the URL above failed, the mvn protocol needs to be installed.
         // not advice create a wrap URLStreamHandlerFactory class now
         try {
-            final Field factoryField = URL.class.getDeclaredField("factory");
```
sgandon, line 94: Why did you remove this check ?

daikon/src/test/java/org/talend/daikon/number/BigDecimalFormatterTest.java
wwang-talend: if run it on java 17 runtime, the junit will fail as the "space" char is not "160", but another one.

daikon/src/main/java/org/talend/daikon/runtime/RuntimeUtil.java
wwang-talend: since java 17, default don't allow call reflect on java package to access the private info, so if not remove this, will throw exception on java 17 here. And check the code, seems this is an old usage, no need to do that now.

daikon/src/test/java/org/talend/daikon/number/BigDecimalFormatterTest.java
coheigea: @wwang-talend Is it possible instead of 160 to refer to a constant that will work with both JDK versions?

daikon/src/main/java/org/talend/daikon/runtime/RuntimeUtil.java
sgandon: ok

daikon/src/test/java/org/talend/daikon/number/BigDecimalFormatterTest.java
sgandon: @wwang-talend any answer on Colm proposal ?

wwang-talend: i will check it

wwang-talend: and now the regex pattern assert also works for both java17/11/8

sgandon (approved):
LGTM

sgandon (dismissed):
LGTM

sgandon (approved):
LGTM

PR #786, opened 2021-11-22
Removing daikon-spring-mongo from the build

coheigea:
Remove daikon-spring-mongo from the build. It doesn't build any more because the Spring maven repo has disallowed anonymous downloads, and I can't find the dependencies on Maven Central.


sgandon (approved):
LGTM

PR #808, opened 2022-01-25
feat(APIC-774): add ECS fields logging

ngrange-talend:
 - Add a way to add ECS fields into the log
 - ECS version: 1.8
 - fields added:
     - labels
     - event 

See https://github.com/Talend/apimgmt-misc/blob/master_POC_daikon/study/daikon/README.md

Below an example to understand how to use the new feature:
```
final EventField fieldEvent = Field.eventFieldBuilder()
        .action("API-configuration-creation")
        .category(EventField.Category.CONFIGURATION)
        .kind(EventField.Kind.EVENT)
        .outcome(EventField.Outcome.SUCCESS)
        .reason("A user created a new API on a Dataset.")
        .build();

final LabelsField fieldLabels = Field.labelsFieldBuilder()
        .addLabel("dataset-id", "{DATASET_ID}")
        .addLabel("api-configuration-id", "{API_CONFIGURATION_ID}")
        .build();

final FieldsMarker ecsFields = FieldsMarker.newFieldsMarker()
        .defineEvent(fieldEvent)
        .defineLabels(fieldLabels)
        .build();

log.info(ecsFields, "Api created successfully.");
```
 

**Link to the JIRA issue**

 

anicolas:
Awesome work @ngrange-talend  👏 

May I suggest some small changes to the API?

I think it's important to stick to the ECS vocabulary here: https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html

> ECS defines multiple groups of related fields. They are called "field sets".

So in my opinion, one should define objects like `EventFieldSet`, `LabelsFieldSet`, ... to regroup fields.

Also I think that the "root" object should be called `EcsFieldsMarker`, WDYT? On the builder for this object, I find it a bit weird to use `defineXXX` to set attributes as it's not the convention used in other builders.

Here is a quick updated API proposal:

```java
final EventFieldSet eventFieldSet = EcsFieldSet.eventFieldSetBuilder()
        .action("API-configuration-creation")
        .category(EventField.Category.CONFIGURATION)
        .kind(EventField.Kind.EVENT)
        .outcome(EventField.Outcome.SUCCESS)
        .reason("A user created a new API on a Dataset.")
        .build();

final LabelsFieldSet labelsFieldSet = EcsFieldSet.labelsFieldSetBuilder()
        .addLabel("dataset-id", "{DATASET_ID}") // <-- Here using addXXX makes sense since it can be called multiple times
        .addLabel("api-configuration-id", "{API_CONFIGURATION_ID}")
        .build();

final EcsFieldsMarker ecsFieldsMarker = EcsFieldsMarker.builder()
        .eventFields(eventFieldSet) // <-- Not 100% sure about those :)
        .labelFields(labelsFieldSet)
        .build();

log.info(ecsFieldsMarker, "Api created successfully.");
```

I'm still a bit puzzled on:

```java
final EventFieldSet eventFieldSet = EcsFieldSet.eventFieldSetBuilder()
        .action("API-configuration-creation")
        .category(EventField.Category.CONFIGURATION)
        // ....
```

vs

```java
final EventFieldSet eventFieldSet = EventFieldSet.builder()
        .action("API-configuration-creation")
        .category(EventField.Category.CONFIGURATION)
        // ....
```

I think these are not exclusive, and the former has the advantage of added discoverability since the builders for all field sets can be found as static methods on the `EcsFieldSet` class 🤔 

Would be nice to have some other feedback regarding this API 😉

ngrange-talend:
I agree with you @anicolas, thanks for the suggestion!
I updated the code :+1: 

Also, I removed the builders in parent class `EcsFieldSet`. Indeed it allowed to to something like: `EventFieldSet.labelsFieldSetBuilder()...` because `LabelsFieldSet` and `EventFieldSet` extends `EcsFieldSet`

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/EcsFieldsMarker.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    public boolean contains(final String name) {
+        if (name == null) {
+            return false;
+        }
+        if (ECS_FIELDS_MARKER_NAME.equals(name)) {
```
pierCo, line 110:
```suggestion
        if (getName().equals(name)) {
```

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/field/EventFieldSet.java
```diff
@@ -0,0 +1,245 @@
...
+        USER("user");
+
+        public final String value;
+
+        private Type(final String value) {
+            this.value = value;
+        }
+    }
+
+    public static final class EventFieldSetBuilder {
```
anicolas, line 108: Can simply be `Builder` (`EventFieldSet.Builder`)

daikon-logging/logging-event-layout/src/main/java/org/talend/daikon/logging/ecs/field/EventFieldSet.java
```diff
@@ -0,0 +1,245 @@
...
+            fields.put(EcsFields.EVENT_REFERENCE, reference);
+            return this;
+        }
+
+        public EventFieldSetBuilder riskScore(final Float riskScore) {
+            fields.put(EcsFields.EVENT_RISK_SCORE, riskScore);
+            return this;
+        }
+
+        public EventFieldSetBuilder riskScoreNorm(final Float riskScoreNorm) {
```
anicolas, line 206: (Commenting here but it's valid for all attributes)

Is it valid to have `null` values here?

If yes, one should probably add a `@Nullable` annotation

ngrange-talend: It is matter of choice I guess... I propose to define the fields as not nullable. Then I changed `Float` and `Long` for `float` and `long` :+1:

ngrange-talend: In fact I changed my opinion on it... Sometimes we don't really know the content of the variable and I think what we absolutely want to avoid is to throw an exception when logging! To handle this, I'll make all fields nullable :+1:

jehervy (dismissed):
Great job ! 
It looks good to me 👌

jehervy (approved):

PR #830, opened 2022-03-30
TDKN-336 - Update Atlassian JIRA client to 5.2.x

coheigea:
This task is to update Atlassian Jira Client in daikon-documentation to 5.2.x, to remove a dependency on Spring 4.


sgandon (approved):
LGTM

PR #854, opened 2022-06-08
There's no need to manually override spring cloud function context an…

coheigea:
…y more since Spring Boot updates


sgandon (approved):
LGTM

PR #856, opened 2022-06-08
Revert "Security: Switch semgrep PR scan to Talend workflow"

Gby56:
Reverts Talend/daikon#855


Gby56:
Reverting, removed the SARIF output that was failing all the time, because the SARIF parser from CodeQL is broken and not supporting new fields from the SARIF standard...

coheigea (approved):

PR #857, opened 2022-06-13
feat: added daikon-spring-license module

IevgenTarasov:

IAM-onprem license control code requires frequent calls from Apps to IAM to signal that the user session is still alive. This is done with code in oidc-client, which introduces a filter which in turn makes these requests to IAM - so all this is happening transparently for the Apps.

With retirement of oidc-client we need to put this code somewhere, where it can be reused by the Apps going on. Since it's on-prem specific, it makes no sense to put it in more generic daikon-spring-auth. Instead we are creating a separate module for it, so that only Apps that have onprem part can pick it up.

This code was copied from oidc-client as is and so might require a slight adjustment, in particular with regards to integration with the App itself (it relies no on Zuul, which might not be used by a given App).
 
**Link to the JIRA issue**

 

IevgenTarasov:
@sgandon the apps in question are basically TDP/TDS/MDM
And I will add a readme, you are absolutely right

PierreReliquet (dismissed):

PierreReliquet (approved):

PierreReliquet (approved):

PR #858, opened 2022-06-14
Remove optional oauth2 dependency from spring-security-oauth2-autocon…

coheigea:
…figure

This optional dependency was dragging in some old Spring OAuth2 module with CVEs, and doesn't seem to be needed.


sgandon (approved):
LGTM

PR #865, opened 2022-07-07
Updating Spring Boot to 2.6.9

coheigea:
Also remove version for Snakeyaml and pick it up from Spring Boot instead.


sgandon (approved):
LGTM

PR #872, opened 2022-07-27
feat(TDP-10831/TQL): Allow conversion of a TQL AST to DSEL

quilicicf:

In TDP, we need to convert a TQL filter to DSEL after processing the TQL filter.

The processing of the TQL filter must be done on the AST!
 
Allow the converter to accept the AST directly.
 
**Link to the JIRA issue**

[TDP-10831](https://jira.talendforge.org/browse/TDP-10831)
 
- [x] Tests for the changes have been added (for bug fixes / features, coverage should be over 75% in the new code) => no need, the new method is called by an old one that's tested already 
- [x] Docs have been added / updated (for bug fixes / features) => no need, very small change, JavaDocs up-to-date
- [x] Related design / discussions / pages (not in JIRA), if any, are all linked or available in the Pull Request => there were none, pretty straight-forward stuff
 

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/DSELConverter.java
```diff
@@ -10,25 +10,34 @@
...
+        return convert(tqlExpression);
+    }
 
+    /**
+     * Utility method to convert a TQL query to a DSEL query.
+     *
+     * @param tqlQuery TQL query
+     * @return DSEL ELNode ready to serve for DSEL interpreter
+     */
+    public static ELNode convert(final Expression tqlQuery) throws TqlException {
```
abenali-talend, line 32: TqlElement is not more suitable than Expression ?
```suggestion
    public static ELNode convert(final TqlElement tqlElement) throws TqlException {
```

quilicicf: I'm not sure, the old method was casting the result of `Tql#parse` in an `Expression` so I kept it as is. 
Since `Tql#parse` returns an `Expression` I've got the feeling that it's the right level in the class tree but I'll let more experienced people confirm.

abenali-talend (dismissed):

abenali-talend (approved):

PR #879, opened 2022-08-12
feat(TDP-11256): RunConv-Back DSEL->TQL converter

fway87:
feat(TDP-11256): RunConv-Back DSEL->TQL converter

* Adding a converter of DSEL to TQL
* Adding tests for this converter
* Refactoring some code of the DSEL->TQL converter and its tests.

**Link to the JIRA issue**
https://jira.talendforge.org/browse/TDP-11256


daikon-tql/daikon-tql-core/src/main/java/org/talend/tql/api/TqlBuilder.java
```diff
@@ -457,6 +458,24 @@ public static Expression isValid(String fieldname) {
...
+
+        // Creating simple isNull expression
+        FieldIsNullExpression fieldIsNullExpression = new FieldIsNullExpression(new FieldReference(fieldname));
+        Expression[] fieldIsNullExpressions = new Expression[] { fieldIsNullExpression };
+
+        // Adding it to a new AST
+        AndExpression andExpression = new AndExpression(fieldIsNullExpressions);
+        return new OrExpression(andExpression);
+
+    }
```
quilicicf, line 477: This `OR(AND(EXPRESSION))` wrapping is indeed pretty weird. 
Did you get to the bottom of this? I'm curious about why we are adding all that wrapping.

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlConverter.java
```diff
@@ -0,0 +1,51 @@
...
+        return tqlExpression;
+    }
+
+    /**
+     * Utility class to print the ELNode as a tree
+     *
+     * @param tree ELNode to print
+     */
+    public static void printELNode(ELNode tree) {
+        Logger LOGGER = LoggerFactory.getLogger(TqlToDselVisitor.class);
```
quilicicf, line 43: I'm surprised that we use a logger linked to another class and re-instantiate it each time this method is called :thinking: 
This looks like a botched copy-paste since the class it `TqlToDselVisitor` whereas this class uses `DselToTqlVisitor`.

We should at least fix the class this logger is attached to, at best create a single static logger linked to `DselToTqlConverter` IMO.

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+        return this.visitChildren(rootNode);
+    }
+
+    protected Expression visitChildren(ELNode node) {
+        return this.visitNodes(node.getChildren());
+    }
+
+    protected Expression visitNodes(List<ELNode> nodes) {
+        final ELNode elNode = nodes.get(0);
+        return elNode.accept(this);
```
quilicicf, line 32: I am quite puzzled by this :thinking: 
We're passing a list but only processing the first argument, do we have a guarantee that the list always contains one and only one element? 

## If this is done on purpose

For specific use-cases where this guarantee stands, I'd make this crystal-clear to avoid confusions. 

Something like:
* having a single method `visitSingleChild(List<ELNode> nodes)` (private if possible) 
* making it check that there is a single element in the list and throw with an explicit error if that's not the case
* calling it directly from the places where the guarantee isn't hidden by a layer of abstraction. Like in `visitRoot`, calling directly `visitSingleChild(rootNode.getChildren())`
* commenting the code to explain why the guarantee stands because it's a bit unsettling as is

## If this is not done on purpose

I'd either:
* make sure the limitations of this converter are known (in Javadoc?) and a JIRA is created explaining the shortcomings of the current implementation and how to extend it later. Also, I'd make the conversion fail hard with an explicit message.
* implement the remaining code paths for when there are multiple children

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    @Override
+    public Expression visitLiteral(ELNode elNode) {
+        throw new IllegalStateException("Unsupported literal: " + elNode.getImage());
+    }
+
+    @Override
+    public Expression visitArray(ELNode elNode) {
+        throw new IllegalStateException("Unsupported array: " + elNode.getImage());
```
quilicicf, line 47: I'm surprised that this is not useful for DSEL -> TQL conversions :thinking: 
Aren't array nodes used in `between` clauses for example? :thinking:

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    @Override
+    public Expression visitCompOp(ELNode elNode) {
+        final String fieldName = elNode.getChild(0).getImage();
+        final String secondChildValue = elNode.getChild(1).getImage();
+        final ELNodeType secondChildType = elNode.getChild(1).getType();
```
quilicicf, line 54: Do we have the guarantee that the first field is always a field?
Shouldn't we assert that both values are not literals (in which case it doesn't make sense to compare them since the result doesn't depend on the items in the filtered collection).

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+
+        switch (elNode.getType()) {
+        case EQUAL:
+            if (secondChildType.equals(ELNodeType.STRING_LITERAL)) {
+                return TqlBuilder.eq(fieldName, secondChildValue);
+            } else if (secondChildType.equals(ELNodeType.HPATH)) {
+                return TqlBuilder.eq(fieldName, secondChildValue);
```
quilicicf, line 61: Outchie, a literal and a field name are passed exactly the same way, this doesn't feel right. 
I have the feeling that we shouldn't use the TqlBuilder here but lower-level APIs.
Something like:

```java
final ELNode leftParameter = elNode.getChild(0);
final ELNode rightParameter = elNode.getChild(1);

if (isLiteral(leftParameter) && isLiteral(rightParameter)) {
  throw IllegalArgumentException("Comparing two literals returns the same result for all items, bla bla");
}

final TqlElement leftTqlElement = isLiteral(leftParameter) ? visitLiteral(leftParameter) : visitHPath(leftParameter);
final TqlElement rightTqlElement = isLiteral(rightParameter) ? visitLiteral(rightParameter) : visitHPath(rightParameter);

switch (elNode.getType()) {
  case EQUAL: new ComparisonExpression(ComparisonOperator.Enum.EQ, leftTqlElement, rightTqlElement);
  ...
}
```

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    @Override
+    public Expression visitNot(ELNode elNode) {
+        return TqlBuilder.not(elNode.getChild(0).accept(this));
+    }
+
+    @Override
+    public Expression visitHPath(ELNode elNode) {
+        throw new IllegalStateException("Unsupported HPath: " + elNode.getImage());
```
quilicicf, line 151: Same question as for literals.

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+            throw new IllegalStateException("Unsupported function: " + elNode.getImage());
+        }
+    }
+
+    @Override
+    public Expression visitAssignment(ELNode elNode) {
+        /*
+         * `field_name=123` is considered as an assignment, despite this is the following which should be an assignment:
+         * `let field_name=123`. Finally we decided to consider this case as an equality.
+         */
```
quilicicf, line 195: I don't understand this comment :thinking: 
Also, can it be done as Javadocs so that it's shown by IDEs from callers without needing to enter this class?

It seems rather dangerous to me to consider something that should not be an equality as one. I'd rather throw than generate ambiguous behavior. But since I'm not sure I understand the comment, I'll wait for a rephrasing to make sure I'm not spewing non-sense.

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+        /*
+         * DSEL supports this since V1.1.0 but not TQL
+         */
+        throw new IllegalStateException("Unsupported conditional expression : " + elNode.getImage());
+    }
+
+    private Expression[] buildExpressionsArray(ELNode elNode) {
+        final Expression[] expressions = new Expression[elNode.getChildren().size()];
+        IntStream.range(0, elNode.getChildren().size()).forEach(i -> expressions[i] = elNode.getChildren().get(i).accept(this));
+        return expressions;
```
quilicicf, line 218: _Nitpick_
This is a bit hard to read, you can simplify it like this:

```suggestion
        return elNode.getChildren()
                .stream()
                .map(child -> child.accept(this))
                .toArray(Expression[]::new);
```

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,262 @@
...
+        switch (elNode.getChild(1).getType()) {
+        case INTEGER_LITERAL:
+            return TqlBuilder.between(fieldName, Integer.parseInt(secondChildValue), Integer.parseInt(thirdChildValue));
+        case DOUBLE_LITERAL:
+            return TqlBuilder.between(fieldName, Double.parseDouble(secondChildValue), Double.parseDouble(thirdChildValue));
+        case STRING_LITERAL:
+            return TqlBuilder.between(fieldName, secondChildValue, thirdChildValue);
+        default:
+            throw new IllegalStateException("Unsupported ELNodeType: " + elNode.getChild(1).getType());
+        }
```
quilicicf, line 235: Same comment as above. 
I'd rather use the lower-level API and `visitLiteral` to keep the literal conversion in a single location rather than duplicating.

daikon-tql/daikon-tql-core/src/main/java/org/talend/tql/api/TqlBuilder.java
fway87: I'm totally agree with you !
That produce lot of "noises", especially for the complex expressions.
That was a question I posed to the team on August 19: https://talend.slack.com/archives/C03ACDAPBM2/p1660141452023629 and had the answer from Aurelien Fourmi : the reason is about the TQL parser (https://github.com/Talend/daikon/blob/master/daikon-tql/daikon-tql-core/src/main/antlr4/TqlParser.g4) where the root expression must be composed of :
```
expression : orExpression;
orExpression : andExpression (OR andExpression)*;
andExpression : atom (AND atom)*;
```

So I need to build any expression with the couple "OR(AND())"

quilicicf: That's not what I read here :thinking: 
My understanding of the grammar is that:
* an `expression` must be an `orExpression`
* an `orExpression` can either be an `andExpression` or any number of `andExpression`s joined by the keyword `OR`
* an `andExpression` can either be an `atom` or any number of `atom`s joined by the keyword `AND`

Which means that an `expression` can be a simple `orExpression` (without `OR` suffix) that is a simple `andExpression` (without `AND` suffix), meaning it can be directly an `atom`.

I may be mistaken but I don't believe so :thinking: 

@gmzoughi can you please confirm? :pray:

quilicicf: Seen with Ghassen. 

He can't recall exactly but he thinks something was breaking somewhere if the wrapping wasn't done. 

Since this is blurry, I would suggest the following.

* remove the wrapping only on the `TqlBuilder#not` method for now and comment why it doesn't work like other `TqlBuilder` methods
* assess what works and what doesn't only on that method

## If nothing breaks

We can remove the wrapping everywhere else later, we get rid of something we don't understand in our own code, full benefits.

## Otherwise

We only have broken a single (new!) method and not broken anything that already exists, this means.

* We'll understand what breaks and why
* We can discuss a fix or just wrap again and heavily comment why, depending on what goes wrong

WDYT?

gmzoughi: I understand and it's logical, but I'm not sure it won't break elsewhere. Here's an example
image
The first implementation seems correct for me and from TDS experience, this has no impact on the performances, parsing the TQL tree is quite quick.
If you want to improve the existent you're welcome as long as there are no regressions.
Please be aware that TDS and TDP for example are highly relying on TQL language.

quilicicf: Hum, nope. 
Seen with Ghassen again, removing the wrapping hack is way more complicated than it seems.
It would mean re-writing either the grammar, `TqlBuilder` or both. 
This is not worth the effort and not the purpose of this PR so let's forget it for now. 

We should ideally contribute later to fix it though. The wrapping nodes are polluting the filters a lot and make tests way harder to read.

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -164,22 +138,22 @@ public Expression visitFunctionCall(ELNode elNode) {
...
                     : TqlBuilder.contains(fieldName, value);
-        case BETWEEN: {
+        case org.talend.maplang.el.interpreter.impl.function.builtin.Between.NAME: {
             return visitBetweenFunction(elNode);
         }
-        case IN: {
+        case BUILTIN_FUNCTION_IN: {
             return visitInFunction(elNode);
         }
         default:
```
quilicicf, line 181: This seems sub-optimal, don't we have a foolproof way of finding out what function call a node represents without having to extract our own constant? :thinking:

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -1,35 +1,36 @@
...
-
-    protected Expression visitNodes(List<ELNode> nodes) {
-        final ELNode elNode = nodes.get(0);
-        return elNode.accept(this);
+        Expression value = null;
+        for (int i = 0; i < node.getChildren().size(); i++) {
+            ELNode child = node.getChildren().get(i);
+            value = child.accept(this);
+        }
+        return value;
```
quilicicf, line 34: That's not what I had in mind for this. 
The  previous version was only handling the first item, this one only handles the last but converts everything. 
It's not better more less efficient :thinking: 

Are we sure that this method is only supposed to get nodes with a single child or not?

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,263 @@
...
+
+            if (!isLiteral(child.getType().name())) {
+                throw new IllegalArgumentException(
+                        "At least one of the parameters is not a literal, whereas the in function can only accept literals");
+            }
+
+            literalValues[i - 1] = (LiteralValue) visitLiteral(child);
+        }
+
+        return new FieldInExpression(field, literalValues);
```
quilicicf, line 252: The index shifting in this code is a bit mind-bending. 
Can we simplify this?

```suggestion
    private TqlElement visitInFunction(ELNode elNode) {
        final TqlElement field = elNode.getChild(0).accept(this);
        List<TqlElement> candidateValues = elNode.getChildren()
            .stream()
            .skip(1) // The first child is the field, the rest is the collection its value must belong to
            .map(child -> {
                if (!isLiteral(child.getType().name())) {
                    throw new IllegalArgumentException(
                            "At least one of the parameters is not a literal, whereas the in function can only accept literals");
                }

                return (LiteralValue) visitLiteral(child);
            })
            .collect(Collectors.toList());
            
        return new FieldInExpression(field, candidateValues);
```

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    protected boolean isLiteral(String tokenType) {
+        for (String lit_type : DSELConstants.LITERAL_TYPES) {
+            if (lit_type.equals(tokenType)) {
+                return true;
+            }
+        }
+        return false;
+    }
```
quilicicf, line 262: There has to be a better way to do this :thinking: 
It's super weird that we need to get the `String` representation of an enum element to compare in a list of strings.
Side note: using `Set#contains` is faster than iterating or using `List#contains`. 
That's not too bad here given the size of the list but these performance holes tend to add up and the final product can become sluggish if this is not controlled.

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -139,12 +145,21 @@ public TqlElement visitNot(ELNode elNode) {
...
+
         return TqlBuilder.not((Expression) elNode.getChild(0).accept(this));
     }
 
     @Override
     public TqlElement visitHPath(ELNode elNode) {
-        throw new IllegalStateException("Unsupported HPath: " + elNode.getImage()); // TODO Dev here and add tests.
+        if (isLiteral(elNode.getType().name()) || ELNodeType.BOOLEAN_LITERAL.name().equals(elNode.getType().name())) {
+            return visitLiteral(elNode);
+        }
```
quilicicf, line 160: I think a HPath cannot be a literal because it represents a field from the filtered collection's items.
:thinking:

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
fway87: I just did the change

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
fway87: This is optimization and seems to be overkill for this switch usae.

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
fway87: The array is not yet supported by the TQL language.
To support them, this will require to modify the TQL parser (TqlParser.g4 file) as well as other files such as the 2 converters TQL->DSEL and DSEL->TQL, test files for these converters and the TQL language.
This seems to me clearly outside the scope of the ticket, so I'm in favor of not supporting arrays for DSEL->TQL conversion.

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -0,0 +1,286 @@
...
+         */
+        throw new IllegalStateException("Unsupported Variable: " + elNode.getImage());
+    }
+
+    @Override
+    public TqlElement visitFunctionCall(ELNode elNode) {
+        final String fieldName = elNode.getChild(0).getImage();
+
+        switch (elNode.getImage()) {
+        case org.talend.maplang.el.interpreter.impl.function.builtin.IsEmpty.NAME:
```
pgautier83, line 180:
```suggestion
        case IsEmpty.NAME:
```
And same for all case below to be more readable

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
```diff
@@ -33,7 +33,13 @@ protected TqlElement visitChildren(ELNode node) {
         TqlElement value = null;
         for (int i = 0; i < node.getChildren().size(); i++) {
             ELNode child = node.getChildren().get(i);
-            value = child.accept(this);
+
+            if (i == 0) {
```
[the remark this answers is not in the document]
fway87: This code is a big reason why I was hoping to get an answer from Cyprien in one of the threads in this PR. 
I just quoted you in the relevant thread.

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
fway87: Good idea, thanks, I just did the changes

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
fway87: I just did a simplification which uses the idea of Cyprien :)

daikon-tql/daikon-tql-dsel/src/main/java/org/talend/tqldsel/dseltotql/DselToTqlVisitor.java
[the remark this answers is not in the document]
fway87: This could be done more later, because unfortunately it's not yet clear for me.

abenali-talend (commented):
Most of requirements seems ok, we can see optimisations to do later

daikon-tql/daikon-tql-core/src/main/java/org/talend/tql/model/FieldMatchesRegex.java
```diff
@@ -26,6 +26,11 @@ public String toString() {
         return "FieldMatchesRegex{" + "field='" + field + '\'' + ", regex='" + regex + '\'' + '}';
     }
 
+    @Override
+    public String toQueryString() {
+        return field.toQueryString() + " " + regex;
```
pgautier83, line 31: where did you find this syntax for regex?

pgautier83: There is no quote or ""?

pgautier83 (dismissed):
Just one question

daikon-tql/daikon-tql-core/src/main/java/org/talend/tql/model/FieldMatchesRegex.java
fway87: My bad, the expression must contain a tild char "~". I just committed a fix and enable the unit test to test the Matches TQL function :)
For the quotes, it's already in the regex attribute

pgautier83 (dismissed):

abenali-talend (approved):

PR #881, opened 2022-08-30
Overriding Nimbus version to fix CVE

coheigea:
Overriding Nimbus (9.14 -> 9.22) version to fix CVE (CVE-2022-21449)
Once we update to Spring Boot 2.7.x we can remove this override.


sgandon (approved):
LGTM

PR #899, opened 2022-10-13
Clesaec/tdkn 343 tree

clesaec:
Key/Value big collection that are hard to keep in memory.
 
 Key/Value collection stored on disk.

**Link to the JIRA issue**
https://jira.talendforge.org/browse/TDKN-343
 

sgandon (dismissed):
LGTM

daikon-collections/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+To build module, simply run `mvn install`
+
+```bash
+$> mvn install
+```
+
+## Usage
+
+This library offers [AVL Tree](https://www.programiz.com/dsa/avl-tree) on disk (to search values by keys).
```
amaumont, line 18: You could probably add more details on how it works, when it writes data on disk.

What could be the eventual improvements/optimizations ?
For example, and if it seems relevant compared to the IO latency:
- write on disk only when the tree size exceeds the limit ?
- OR write on disk asynchronously using a queue of operations ?

Give the precision that the POJO have to implement `Comparable`.

amaumont (approved):
Nothing special to change in the code, but maybe more details on how it works effectively, and what potential improvements could be done.

daikon-collections/README.md
clesaec: Details added.For Comparable, it's only needed for Key Class.
_write on disk only when the tree size exceeds the limit ?_ => all tree is write on disk; it's the memory usage once the file is built that can vary. For AvlTreeBuilder (that write on isk), i think it's more on file system side to deal with buffer if necessary.
_write on disk asynchronously using a queue of operations ?_ Same, on file system side, but can be very  for RandomAccess file (data are not just added to end of file, but also modified inside (for pointer to left & right child)

amaumont: Thank you, just let me add a not related proposal in an other new comment.

amaumont:
I would eventually need an iterator to read the tree in the natural order, what's your opinion ?

amaumont (approved):
LGTM

PR #913, opened 2022-10-25
Test daikon-spring-examples and remove wildcard actuator endpoints

coheigea:
Remove wildcard actuator endpoints and just use the minimum needed, to prevent someone copying the files and inadvertently exposing sensitive endpoints.


sgandon (approved):
LGTM

PR #920, opened 2022-11-08
Aligning Jackson versions in logging-kafka-interceptor

coheigea:
Aligning Jackson versions in logging-kafka-interceptor


sgandon (approved):
LGTM

PR #921, opened 2022-11-08
Aligning Jackson versions in logging-request-interceptor

coheigea:
Aligning Jackson versions in logging-request-interceptor


sgandon (approved):
LGTM

PR #936, opened 2022-12-14
Updating Kafka clients from 2.8.1 to 2.8.2 to fix CVE-2022-34917

coheigea:
kafka-clients is vulnerable to denial of service. An attacker can crash the application through the OutOfMemoryException in the `readArray` function of `ByteBufferAccessor.java` by providing large amounts of memory on brokers.


sgandon (approved):
LGTM

PR #972, opened 2023-05-23
TPSVC-20004: make correlationId configuration conditional

vgagara-talend:
When tracing is disabled for app (and OTEL tracer is used) then no Tracer is configured by spring. Thus it leads to error when configuring CorrelationIdProvider/CorrelationIdSetter.
 
Make configuration conditional. Use the same condition as in Spring's ZipkinAutoConfiguration. 

**Link to the JIRA issue**

 

daikon-messages/messages-model-spring-support/src/main/java/org/talend/daikon/messages/spring/producer/CorrelationIdProvidersConfiguration.java
```diff
@@ -10,30 +10,32 @@
...
 @AutoConfigureBefore({ DefaultProducerProvidersConfiguration.class })
-public class SpringSleuthProvidersConfiguration {
+public class CorrelationIdProvidersConfiguration {
 
     @Value("${org.talend.daikon.messages.spring.producer.sleuth.defaultSpanName:MESSAGING_DEFAULT_SPAN")
     private String defaultSpanName;
 
     @Bean
-    public CorrelationIdProvider sleuthCorrelationIdProvider(@Autowired final Tracer tracer) {
+    @ConditionalOnProperty(prefix = "management.tracing", name = "enabled", matchIfMissing = true)
```
fabballe, line 37: Why not adding this conditionalOnProperty on the class level ?

daikon-messages/messages-model-spring-support/src/main/java/org/talend/daikon/messages/spring/consumer/SpringSettersConfiguration.java
```diff
@@ -28,6 +29,7 @@
 public class SpringSettersConfiguration {
 
     @Bean
+    @ConditionalOnProperty(prefix = "management.tracing", name = "enabled", matchIfMissing = true)
```
fabballe, line 32: Why not adding this conditionalOnProperty on the class level ?

daikon-messages/messages-model-spring-support/src/main/java/org/talend/daikon/messages/spring/producer/CorrelationIdProvidersConfiguration.java
vgagara-talend: in this class we have another one "dummy" provider which is used when tracing disabled. I think if we put it in class level both will be ignored.

daikon-messages/messages-model-spring-support/src/main/java/org/talend/daikon/messages/spring/consumer/SpringSettersConfiguration.java
vgagara-talend: Yes, in this case it can be in class level (just copied it from previous case). I will move it to class level.

fabballe (approved):

PR #974, opened 2023-05-24
fix(schema): Allow serialization/de-serialization without a custom mapper

quilicicf:

People that need to de-serialize the `DatasetSchema` currently need to use a special `ObjectMapper` which can the following issues:

* Some (most?) web frameworks instantiate the mapper internally and it's not trivial to change it. Sometimes, the settings on the current mapper won't even fit the rest of the objects the web server need to de-serialize and people will need to fine-tune which mapper to use for which class. This can easily become a pain.
* Most de-serialization specificities can be dealt with using annotations. People familiar with Jackson that just need to de-serialize the schema (my case, exactly) might be tempted to skip the doc (or not even look for it) because usually the annotations are enough (exxactly the pitfall I fell into). This is not too bad (I found the doc relatively fast) but still a waste of time (I asked two people for help in the process) that can be avoided.

[Original discussion (Slack)](https://talend.slack.com/archives/CBA9JCRTR/p1684341172857029). Should the PR be accepted, I'll report all the information in the JIRA issue I'll create.
 
Get rid of the custom object mapper by implementing the specificities of the de-serialization entirely via annotations.

The most complex specificity was the deserialization of the `"null"` type, combined with the polymorphism of the type and the fact that it's in a list.

[Here's more information about why that's not trivial](https://github.com/FasterXML/jackson-databind/issues/1654#issuecomment-307853268).

To fix this, the solution was to implement our own serializer/de-serializer for the __types list__, plugged to the POJO by annotation. This adds some boilerplate but:

* it's rather simple code that shouldn't break
* it's auto-documented (annotation + class Javadoc)
* it's tested (JUnit tests)

So I think it is a bargain.
 
**Link to the JIRA issue**

I'll create the JIRA issue if you think this adds value and should be added to main.
 
- [x] Related design / discussions / pages (not in JIRA), if any, are all linked or available in the Pull Request
 
There's no reason why this would break anything, unless we remove the previous mapper now (in which case we'll need to ask people to use any standard mapper instead).

I don't know the policy for deprecation in this repository, so I haven't removed the mapper yet and will wait for instructions. We can remove it directly or deprecate it for example.

> :information_source: : The `NullTypeStringSerializer` is no longer needed and was part of a specific unit test that has been made irrelevant with the new implementation.
> 
> Depdending on what we decide to do about the previous mapper, we'll need to remove the deserializer and the test now or later.


fabballe (commented):
Nice job.

However you should also update DeserializerSerializeNullTest which use a custom objectMapper too

quilicicf:
Yeah, I talked about the `DeserializerSerializeNullTest` in my PR description actually :sweat_smile:.

If you are OK with going full annotation style, I think this PR needs to include the following changes:

* Remove mention of the specific custom mapper in the doc (or leave a paragraph to explain it's deprecated/removed)
* Remove the `NullTypeStringDeserializer`
* Remove the `DeserializerSerializeNullTest`

But I need to know first if we should deprecate the mapper or remove it immediately to take action :pray: 
*

fabballe:
Right now this module is not used a lot so we can just delete the objectMapper, NullTypeStringDeserializer  and all UT linked to it

fabballe (approved):

